PR #56, opened 2022-11-21
Improve AVL API logging

ITO-PaulHolland:
We are now logging the query params and the time it takes to make a request to the consumer api

*DO NOT MERGE*


ITO-PaulHolland:
Dropping as this has already been deployed

PR #67, opened 2022-11-25
fares validator + logging + ITO changes

surabhika:


transit_odp/api/tests/test_avl_api.py
```diff
@@ -1,3 +1,4 @@
+from datetime import timedelta
```
ITO-PaulHolland, line 1: I would expect this to already be in dev from a main rebase

surabhika: Hi Paul,
One cannot commit directly into dev as the branch is protected hence this is going via a PR.

pritish-ranjan01 (approved):

PR #94, opened 2022-12-12
BODP-5619: Feed weekly score not matching

AndreaIto:
between weekly csv reports and on the dataset page

Replaced round function with integer division

87_table

87_detail

@paul.holland


transit_odp/avl/views/detail.py
```diff
@@ -85,11 +85,9 @@ def get_context_data(self, **kwargs):
             ).first()
```
ITO-PaulHolland, line 85: I wonder if it will be possible to add a unit test that compares the two?
If its too hard or will take ages I'm not into it but if its easy it could be a good way to go.

AndreaIto: Ok, I'll do it!

surabhika:
Screenshot 2022-12-13 at 15 37 43

The changes are on dev branch now @ITO-PaulHolland. Please trigger the deployment with this PR. Thank you :)

ITO-PaulHolland (approved):

PR #106, opened 2022-12-19
BODP-5614: fixed width short description

AndreaIto:
Fixed the length of the "short description" in the table


ITO-PaulHolland (commented):
Im going to need screenshot. Ill compare to mine but its a minefield to get this stuff right

PR #110, opened 2023-01-04
BODP-5651: Handle null (zero) values when aggregating daily reports and calculating percentage scores

anguskbell:


transit_odp/avl/post_publishing_checks/weekly/tests/test_summary.py
```diff
@@ -442,3 +442,74 @@ def test_ppc_summary(
...
+                    "Notes": "dummy_1",
+                },
+            ]
+        },
+        {
+            "AVLtoTimetable matching summary": [
+                {
+                    "SIRI field": "LineRef",
+                    "TXC match field": "LineName",
+                    "Total vehicleActivities analysed": 0,
```
ITO-PaulHolland, line 475: do one of these need to be an empty string to trigger an NA?

anguskbell: If the total vehicles analysed is 0, then you get a divide-by-zero in the calculation of the two percentage figures, which fills in `nan`.

ITO-PaulHolland: Well in that case, take my approval sir!

ITO-PaulHolland (approved):

PR #113, opened 2023-01-05
fix/BODS-87

pritish-ranjan01:
This PR includes:-

* Fix for failing test cases

### DIFF
--- a/transit_odp/browse/forms.py
+++ b/transit_odp/browse/forms.py
@@ -182,12 +182,11 @@ class FaresSearchFilterForm(GOVUKForm):
         ),
         required=False,
     )
-    is_fares_validator_active = flag_is_active("", "is_fares_validator_active")
-    if is_fares_validator_active:
-        is_fares_compliant = forms.NullBooleanField(
-            required=False,
-            label=_("BODS compliance"),
-        )
+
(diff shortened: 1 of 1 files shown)


transit_odp/browse/forms.py
```diff
@@ -182,12 +182,11 @@ class FaresSearchFilterForm(GOVUKForm):
...
         required=False,
     )
-    is_fares_validator_active = flag_is_active("", "is_fares_validator_active")
-    if is_fares_validator_active:
-        is_fares_compliant = forms.NullBooleanField(
-            required=False,
-            label=_("BODS compliance"),
-        )
+
+    is_fares_compliant = forms.NullBooleanField(
```
ITO-PaulHolland, line 186: This is cool if you dont care about the field being present regardless of the feature flag which makes things simple. If you do care then you can define another form class without the field and add the logic to the get_form_class method in the view.

ITO-PaulHolland (approved):

transit_odp/browse/forms.py
pritish-ranjan01: We will keep a note of that and create a task in our backlog for this.

PR #115, opened 2023-01-06
BODP-5652: datasetpublishing.csv file

AndreaIto:
datasetpublishing.csv file should display 'Undergoing Validation' or 'Unavailable due to dormant feed' statuses when relevant

Added new logic in the column to create the csv.
Even when the ppc score is not available the status has to be updated properly, created a new function that takes care of retrieving a queryset in the AVLDataset, this function is called in the if statement and once is called, it caches the results, this avoid hammering the database, the subsequents calls are dictionary/key calls

Screenshot 2023-01-06 at 11 24 36


transit_odp/site_admin/exports.py
```diff
@@ -232,13 +255,17 @@ class DatasetPublishingCSV(CSVBuilder):
                 round(ppc_overall_score(dataset.organisation_id))
             )
             + "%"
-            if dataset.percent_matching >= 0 and dataset.status != INACTIVE
```
anguskbell, line 235: Same comment as above. Can you split this out into a dedicated function?

transit_odp/site_admin/exports.py
```diff
@@ -118,6 +118,24 @@ def get_ppc_weekly_overall(organisation_id: int) -> float:
...
+                AVLDataset.objects.get_active()
+                .add_avl_compliance_status_cached()
+                .filter(organisation_id=organisation_id)
+            )
+        return avl_compliance_status[organisation_id]
+
+    return get_avl_compliance_status
+
+
+avl_compliance_status_cached_caller = make_avl_compliance_status()
```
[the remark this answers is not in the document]
AndreaIto: Yes It's a closure, but a Python closure, if you look at row 98 there is another closure.
Here I use the same logic. What I'd avoid here is hammering the db.

transit_odp/site_admin/exports.py
```diff
@@ -232,13 +255,17 @@ class DatasetPublishingCSV(CSVBuilder):
...
+            if ppc_overall_score(dataset.organisation_id)
+            and dataset.percent_matching >= NO_PPC_DATA
+            and dataset.status != INACTIVE
             else "",
         ),
         CSVColumn(
             header="Archived matching reports URL",
             accessor=lambda dataset: get_ppc_weekly_overall_url(dataset.organisation_id)
-            if dataset.percent_matching >= 0 and dataset.status != INACTIVE
+            if ppc_overall_score(dataset.organisation_id)
```
anguskbell, line 266: In what circumstances will `ppc_overall_Score` return a falsy value?

transit_odp/site_admin/tests/test_exports.py
```diff
@@ -196,13 +207,27 @@ def test_active_publisher_to_string_last_four_columns(self, client_factory):
...
+            granularity=PPCReportType.WEEKLY,
+            created=today,
+        )
+        PostPublishingCheckReportFactory(
+            dataset=dataset5,
+            vehicle_activities_analysed=0,
+            vehicle_activities_completely_matching=0,
+            granularity=PPCReportType.WEEKLY,
+            created=today,
+        )
```
anguskbell, line 223: The correct check for "undergoing validation" is that no weekly PPC report has been generated yet. It's probably a grey area whether a report with 0 vehicles analysed constitutes "undergoing validation" - maybe it does. Anyway, better to just not generate a report for this dataset.

transit_odp/site_admin/exports.py
AndreaIto: `ppc_overall_score` return a float, all the `if` statement return:
 1. false if is INACTIVE 
 2. when "% AVL to Timetables feed matching score" is:
     - number
     - Undergoing ..
     -  Due to ....
     Has to be true

transit_odp/site_admin/exports.py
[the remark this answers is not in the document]
AndreaIto: It's similar to a decorator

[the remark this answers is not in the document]
AndreaIto: Yeah sure not a problem  to initialize in the constructor of the `DatasetPublishingCSV`, 
but since this is a bug I'd prefer leave the code already been approved and in `dev` I mean 
I think it's better not touch the code from row `98`, unless you don't want create a new bug story.

transit_odp/site_admin/tests/test_exports.py
anguskbell: I don't see how this has been resolved.

AndreaIto: I removed the granularity that is a constraints

transit_odp/site_admin/exports.py
AndreaIto: The purpose of `ppc_overall_score` is create a `QuerySet`, it returns a `float` but most important it queries the db
calling `.add_avl_compliance_status_cached` in the `AVLDataset`, Unfortunately I can't call in `self.get_queryset` (Dataset).
Probably here it not really clear, but the aim of the `if` is not to see true or false but call the `QuerySet`
I mean I am using the `if` to call a function.
This is what I meant in the first commit message 👆

AndreaIto: "If" you think in the `if` that call is always `True`, "if" you think in the `function` there has been called and it has been called `.add_avl_compliance_status_cached`

anguskbell: Yes, `ppc_overall_score` creates and executes a QuerySet - but it returns the average value and **discards** the rest of the QuerySet. It's not possible for the contents of this QuerySet to be used elsewhere.

Where do you think it's being used?

(If you really do want to retain values in the QuerySet, you would need to make it a member of the CSV builder class.)

transit_odp/site_admin/exports.py
```diff
@@ -126,6 +126,35 @@ def get_ppc_weekly_overall_url(organisation_id: int) -> str:
...
+def latest_matching_url(dataset) -> str:
+    if dataset.percent_matching >= 0 and dataset.status != INACTIVE:
+        return get_ppc_weekly_per_feed_download_report(
+            dataset.organisation_id, dataset.id
+        )
+    return ""
+
+
+def overall_avl_timetables_matching(dataset) -> str:
+    if ppc_overall_score(dataset.organisation_id) and dataset.status != INACTIVE:
```
anguskbell, line 147: There's still this instance of `ppc_overall_score` that you don't need. It doesn't need to be in the `if` statement as it's called on the next line anyway. You're not testing the return value so it's misleading putting it in the `if` clause. Also, for inactive datasets, it's called unnecessarily.

AndreaIto: I think I need it because without I got an error:
`E   TypeError: type NoneType doesn't define __round__ method`

anguskbell: That implies that `ppc_overall_score` is returning `None`. That's unexpected, so what are the circumstances of that? You need to understand why that's happening, then write code to handle it.

AndreaIto: It can be `None` if `AVLDataset` is empty.
The test that I wrote works, because I create the `DatasetFactory` (with `dataset_type=AVLType`
but the that potentially is broken it's the `def test_active_publisher_to_string(self, factory, dataset_type, user_type):`

here the `AVLDataset` is empty and then 
`AVLDataset.objects.get_active()
                .add_avl_compliance_status_cached()
                .exclude(avl_compliance_status_cached__in=[MORE_DATA_NEEDED])
                .filter(organisation_id=organisation_id)
                .add_post_publishing_check_stats()
                .exclude(percent_matching=float(NO_PPC_DATA))
                .aggregate(Avg("percent_matching"))["percent_matching__avg"]`
is `None`.

Calling `round` in `None` raise an error

anguskbell: How about something like this:

```suggestion
    if dataset.status != INACTIVE:
        score = ppc_overall_score(dataset.organisation_id)
        if score is not None:
            return str(round(score)) + "%"
    return ""
```

Preferred because:
* `ppc_overall_score` is not called unnecessarily for inactive datasets
* The expected values returned from `ppc_overall_score` are made explicit

anguskbell (approved):

PR #127, opened 2023-01-13
Updates to ease BODS development on a mac

ITO-PaulHolland:


docker/local/clamav/freshclam.conf
```diff
@@ -5,7 +5,7 @@
 DatabaseDirectory /store
```
ITO-PaulHolland, line 5: @surabhika,
Did you mention that you solved this issue?

PR #131, opened 2023-01-13
NO-TICKET: Fix connecting to Postgres DB inside postgres container

anguskbell:
Also fix CELERY_BROKER_URL and stop using bods.local.

### DIFF
--- a/config/.env.local.template
+++ b/config/.env.local.template
@@ -7,7 +7,7 @@ GID=20
 
 # Django Hosts
 # ------------------------------------------------------------------------------
-DJANGO_PARENT_HOST=bods.local
+DJANGO_PARENT_HOST=localhost
 DJANGO_HOST_PORT=8000
 DJANGO_HOST_SCHEME=http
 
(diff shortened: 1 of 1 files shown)


config/.env.local.template
```diff
@@ -21,6 +21,8 @@ REDIS_URL=redis://redis:6379/0
 
 # Celery
 # ------------------------------------------------------------------------------
+DATABASE_URL=postgis://transit_odp:transit_odp@postgres:5432/transit_odp
```
ITO-PaulHolland, line 24: The DATABASE_URL is generated in the /entrypoint and goesnt need to be defined here. 
Probably should go in a bash_profile or an IDE

anguskbell: It was literally you who gave me this fix! Plus we don't have a bash_profile (and I'm not even using `bash` these days anyway).

I don't know what you want me to do.

ITO-PaulHolland: I think its best if its removed. What the caller sees here will immediately get overridden when the entrypoint is called. 
It is useful to have when running outside a container. Like for when we run `manage.py runserver_plus` Do we have a nodocker template? Maybe its best in there. Or maybe we tell the dev to add it to their environment and they can pick their best way.

anguskbell: Ok, if I omit `DATABASE_URL` from `.env` and restart my containers, then I get an error saying that `DATABASE_URL` is not in the environment.

So it doesn't look like it's getting set from `/entrypoint`. Is that the bug?

anguskbell:
Paul didn't explain what he wanted instead, and has left BODS, so closing.

PR #145, opened 2023-01-18
story/BODS-83: Feature flag for fares data catalogue

ndemenezes:


surabhika (changes requested):
create_guidance_file_string method has guidance for fares data catalogue which also needs to have feature flag check. See line 88 to 93.

surabhika (approved):

PR #147, opened 2023-01-19
BODP-5673: DfT Admin: 'datasetpublishing.csv'

AndreaIto:
file should not list the operator % and archive URL for Timetables and Fares datasets

Capitalized the status as requested from QA

Screenshot 2023-01-19 at 19 30 53


transit_odp/site_admin/exports.py
```diff
@@ -144,7 +144,7 @@ def latest_matching_url(dataset) -> str:
 
 
 def overall_avl_timetables_matching(dataset) -> str:
-    if dataset.status != INACTIVE:
+    if dataset.dataset_type == DatasetType.AVL.value and dataset.status != INACTIVE:
```
anguskbell, line 147: It's not wrong to use `DatasetType.AVL.value`, but `AVLType` is more succinct and that's what it's there for!

anguskbell (approved):

anguskbell (approved):

PR #165, opened 2023-01-25
[CHERRY-PICK] BODP-5683: AVL - Dashboard: General UI

AndreaIto:
Weekly overall score is not updating when the overall percentage is 0%

0 is considered False in if statement, this make clear that has to be different from None.


marcusher-ito:
Will you also need to increment `version.txt` so it deploys properly in test?

PR #168, opened 2023-01-26
BODP-5674: Add SIRI-VM and TXC XML line numbers

anguskbell:


ITO-PaulHolland (approved):
Do you need to change any unit tests? Just a heads up

PR #172, opened 2023-01-26
[CHERRY-PICK]: BODP-5683: PPC E2E AVL - Dashboard: General UI

AndreaIto:
Weekly overall score is not updating when the overall percentage is 0%

The counter in the banner should be an integer.


marcusher-ito:
Hey @AndreaIto - I removed the `|floatformat:"0"` on that line 16 as part of my bugfix for 5684 to fix the rounding error that was happening in that banner metric. Do you mind removing it so it's not added back in?

AndreaIto:
Hey @marcusher-ito I think the banner is showing the float number according with the bug 5683, see the comment, please.

ITO-PaulHolland (approved):

PR #191, opened 2023-02-03
BODP-5713: Now default empty response from API

ITO-PaulHolland:
Before we would raise and this has caused the task to fail when the API returns no-content.

The unit testing has been improved to explicitly test this


ITO-PaulHolland:
Closing this PR as it was merged in error directly to dev branch. 
Caused by stray clicks in Git Kraken, moral of the story is to always use the command line!

PR #193, opened 2023-02-06
fix/BODS:999 handling feature flag behind guidance doc string creation

pritish-ranjan01:


transit_odp/browse/constants.py
```diff
@@ -14,6 +14,40 @@
...
+Data Catalogue
+The data catalogue zip contains a series of CSVs which gives a machine-readable
+overview of all the data that resides in BODS currently.
+
+Note that the data catalogue only covers the data from primary data sources on BODS
+which is timetables data in TransxChange format, bus location data in SIRI-VM format
+and fares data in NeTEx format. Other non-primary data on BODS (e.g disruptions data
+or GTFS converted forms) are not represented on the data catalogue.
+
 The data catalogue zip contains 4 distinct CSVs:
```
surabhika, line 51: There are 5 files in the zip now :)

surabhika (changes requested):

transit_odp/browse/constants.py
pritish-ranjan01: Done

PR #200, opened 2023-02-09
BODS-3 stop retry when OTC has empty response

surabhika:
Hi @ITO-PaulHolland,
Please have a look and suggest if this is fine. 
1. I am handling the empty response at the source which marks the job as success.
2. Remove retry attempts if the response is empty.

### DIFF
--- a/transit_odp/otc/client/otc_client.py
+++ b/transit_odp/otc/client/otc_client.py
@@ -25,7 +25,7 @@ class EmptyResponseException(Exception):
     pass
 
 
-retry_exceptions = (RequestException, EmptyResponseException)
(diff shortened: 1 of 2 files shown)


ITO-PaulHolland:
@surabhika 
Ive fixed the issue here:
https://github.com/department-for-transport-BODS/bods/pull/191

I did this before Ben said he was going to pass it to you. 
I think your code would solve the problem. There were a few things I wanted document before I left the project. I hope this helps

surabhika:
@ITO-PaulHolland  Ahh I see. Thank you :)
You're returning an empty API response! Makes sense

PR #204, opened 2023-02-09
BODP-5734: Changed fields to date fields and added validator

ITO-PaulHolland:
We now check that end date >= today


ITO-PaulHolland:
Screenshot 2023-02-09 at 18 06 40

marcusher-ito (approved):

PR #205, opened 2023-02-09
BODP-5716: Task to advance seasonal service dates automatically

AndreaIto:
We will create a celery task that runs daily and checks for any seasonal services that end in the past. When this happens, the start and end date of the seasonal service will automatically be advanced by a year.

### DIFF
--- a/transit_odp/site_admin/tasks.py
+++ b/transit_odp/site_admin/tasks.py
@@ -4,10 +4,12 @@ from datetime import timedelta
 from celery import shared_task
 from dateutil.relativedelta import relativedelta
 from django.core.files.base import File
+from django.db.models.expressions import F
 from django.utils import timezone
 
 from transit_odp.avl.models import PostPublishingCheckReport, PPCReportType
 from transit_odp.feedback.models import Feedback
+from transit_odp.organisation.models.data import SeasonalService
 from transit_odp.site_admin.constants import (
     ARCHIVE_CATEGORY_FILENAME,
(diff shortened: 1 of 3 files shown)


transit_odp/site_admin/tasks.py
```diff
@@ -168,3 +170,13 @@ def task_delete_unwanted_data():
         .delete()
     )
     logger.info(f"Deleted {deleted} post publishing check weekly report")
+
+
+@shared_task()
+def task_seasonal_service_updated_dates():
+    today = timezone.datetime.today()
+    one_year_deltatime = timedelta(days=365)
```
ITO-PaulHolland, line 178: It might be an idea to use relative delta and 1 year. I think that will take leap years into account.

ITO-PaulHolland:
🎆 Nice query! 🎆

transit_odp/site_admin/tasks.py
AndreaIto: This is really a good idea, but I got this problem:

`ProgrammingError: can't adapt type 'relativedelta'`

`start` and `end` are `DateField` don't accept `relativedelta`.
This can be avoid if :
1. QuerySet
2. sum the dates
3. save the data in db

but this destroy the single QuerySet
@ITO-PaulHolland

ITO-PaulHolland: damn it. leave it like this. I think its ok

PR #210, opened 2023-02-10
BODP-5737: Deleting a seasonal service on 2nd page to Page Not Found

AndreaIto:
Updated the page of the link according with the table.


transit_odp/publish/tests/test_seasonal_service_views.py
```diff
@@ -147,6 +148,39 @@ def test_delete_last_seasonal_service(self, publish_client):
             == self.NUMBER_SEASONAL_SERVICE_PER_ORGANISATION - 1
         )
 
+    def test_transition_page_when_delete_seasonal_service(self, publish_client):
+        SEASONAL_SERVICE_UPGRADE = 6
```
ITO-PaulHolland, line 152: I thought we needed 11 seasonal services to make two pages?

transit_odp/publish/tests/test_seasonal_service_views.py
```diff
@@ -147,6 +148,39 @@ def test_delete_last_seasonal_service(self, publish_client):
...
+            "delete-seasonal-service",
+            host=PUBLISH_HOST,
+            kwargs={"pk1": self.org.id},
+        )
+        publish_client.post(
+            delete_url,
+            data=delete_data,
+        )
+        # the 3rd page no longer exists!
+        response = publish_client.get(self.url + "?page=3")
```
ITO-PaulHolland, line 178: might be an idea to explicitly test the redirect url to make sure it exists and has seasonal services on it

transit_odp/publish/views/timetable/seasonal_services.py
```diff
@@ -177,12 +178,18 @@ def get_form_kwargs(self):
         return kwargs
 
     def get_success_url(self):
+        org_id = self.kwargs["pk1"]
         page = self.request.GET.get("page", None)
+        seasonal_services = SeasonalService.objects.filter(
```
ITO-PaulHolland, line 183: This is already defined in `get_queryset`. I wonder if this means `self.queryset` is now defined or `self.object_list`

transit_odp/publish/tests/test_seasonal_service_views.py
AndreaIto: There are already 15 seasonal services + 6 => 21 seasonal services 3 pages.

transit_odp/publish/views/timetable/seasonal_services.py
AndreaIto: I am in DeleteView there are no `get_queryset`, `self.queryset` and `self.object_list`

ITO-PaulHolland, line 183: You probably want to sort that by end date. In fact this query needs to be the same as the one in the list view.

transit_odp/publish/tests/test_seasonal_service_views.py
AndreaIto: There is no redirect here. The link `self.url + "?page=3"` doesn't exist! and the `follow = True` go straight to 404. 
This pass the test:

`        delete_url = reverse(
            "delete-seasonal-service",
            host=PUBLISH_HOST,
            kwargs={"pk1": self.org.id},
        )
        response = publish_client.post(
            delete_url,
            data=delete_data,
        )
        # redirect to list table
        assert response.status_code == 302
        assert response.url == self.url + "?page=&"
        # the 3rd page no longer exists!
        response = publish_client.get(
            self.url + "?page=3",
            follow=True,
        )
        assert response.status_code == 404

transit_odp/publish/views/timetable/seasonal_services.py
AndreaIto: Hey Paul I can do it if you wish, but it's not relevant here the order of the QuerySet I am passing this qs to the `Paginator`

ITO-PaulHolland (approved):

PR #212, opened 2023-02-14
fix/BODS-14&16: Fixing issue with incorrect UserType counts during publishing flow

pritish-ranjan01:


ndemenezes (approved):

transit_odp/fares/extract.py
```diff
@@ -188,6 +188,14 @@ def _get_count(self, attr):
         lens = [len(lines) for lines in self._attr_from_documents(attr)]
         return sum(lens)
 
+    def _get_user_type_count(self, attr):
+        """Gets the total count of distinct UserType in all documents"""
+        lists_user_types = self._attr_from_documents(attr)
+        distinct_user_types = set(
+            [value for sublist in lists_user_types for value in sublist]
```
surabhika, line 195: You should probably convert each value into lower case and then compare?

pritish-ranjan01: This should not be required as xml schema validation will fail if there is any uppercase value in it..

lxml.etree.XMLSyntaxError: Element '{http://www.netex.org.uk/netex}UserType': [facet 'enumeration'] The value 'Senior' is not an element of the set {'adult', 'child', 'infant', 'senior', 'student', 'youngPerson', 'schoolPupil', 'military', 'disabled', 'disabledCompanion', 'jobSeeker', 'employee', 'animal', 'anyone'}.

surabhika: Ohh makes sense

surabhika (approved):

PR #214, opened 2023-02-14
BODP-5634: Implement SCs Requiring Attention logic

marcusher-ito:
- Added queryset methods to OTC Service manager
- Added queryset methods to TXCFile Attribute manager
- Added queryset methods to Seasonal Services manager
- Spread out collecting of data for services requiring attention to two queries
- Created requiring_attention module
- Added unit tests


transit_odp/organisation/querysets.py
```diff
@@ -1152,6 +1153,32 @@ def get_overall_data_catalogue(self):
...
+        )
+
+    def add_effective_stale_date_end_date(self):
+        """Adds the Effective Stale Date: Operating Period End Date - 42 days."""
+        return self.annotate(
+            effective_stale_date_end_date=ExpressionWrapper(
+                F("operating_period_end_date") - timedelta(days=42),
+                output_field=DateField(),
+            )
+        )
```
anguskbell, line 1172: Wouldn't these functions be quicker if you truncated to a Date (rather than DateTime)? You end up having to truncate later anyway.

transit_odp/organisation/querysets.py
```diff
@@ -1152,6 +1153,32 @@ def get_overall_data_catalogue(self):
     def filter_by_noc_and_line_name(self, noc, line_name):
         return self.filter(national_operator_code=noc, line_names__contains=[line_name])
 
+    def add_effective_stale_date_last_modified_date(self):
+        """Adds the Effective Stale Date: Last Modified Date of file + 1 year."""
+        return self.annotate(
+            effective_stale_date_last_modified_date=ExpressionWrapper(
+                F("modification_datetime") + timedelta(days=365),
```
anguskbell, line 1160: Doesn't handle leap years. I guess you're saying that's an acceptable risk?

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Get a list of dictionaries of live TXCFileAttributes for an organisation
+    with relevant effective staleness dates annotated.
+    """
+    return {
+        txcfa.service_code: txcfa
+        for txcfa in TXCFileAttributes.objects.filter(
+            revision__dataset__organisation_id=org_id
+        ).add_staleness_dates()
+    }
```
anguskbell, line 30: How does this return only the live revisions?

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+    Checks key staleness dates on live TXCFileAttributes and OTC Service to determine
+    if a service is Stale and returns all Stale live services.
+
+    Staleness logic:
+        Staleness Status - Not Stale:
+            Default status for service codes published to BODS
+        Staleness Status - Stale - OTC Variation:
+            If last_modified < OTC Service effective_date
+            AND
+            Today >= effective_stale_date_end_date
```
anguskbell, line 57: Should be Effective stale date due to OTC effective date

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+        Staleness Status - Stale - End Date Passed:
+            If effective_stale_date_end_date (if present) >
```
anguskbell, line 59: should be <

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+                effective_stale_date_last_modified_date
+            AND
+            Today >= effective_stale_date_end_date
+            AND
+            last_modified > OTC Service effective_date
```
anguskbell, line 64: should be >=

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+        Staleness Status - Stale - 12 months old:
+            If effective_stale_date_last_modified_date >
```
anguskbell, line 66: should be <

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+                effective_stale_date_end_date (if present)
+            AND
+            Today >= effective_stale_date_last_modified_date
+            AND
+            last_modified_date > OTC Service effective_date
```
anguskbell, line 71: should be >=

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+    staleness_otc = (
+        effective_date > last_modified
+        and effective_stale_date_otc_effective_date <= today
+    )
+    staleness_end_date = (
+        (
+            effective_stale_date_last_modified_date
+            < effective_stale_date_end_date.date()
```
anguskbell, line 91: Should be >

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+            <= today
+            and effective_date < last_modified
```
anguskbell, line 93: Should be <=

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        if effective_stale_date_end_date
+        else False
+    )
+    staleness_12_months_old = (
+        effective_stale_date_last_modified_date <= today
+        and effective_date < last_modified
```
anguskbell, line 100: Should be <=

transit_odp/publish/requires_attention.py
```diff
@@ -0,0 +1,124 @@
...
+        (
+            effective_stale_date_last_modified_date
+            < effective_stale_date_end_date.date()
+            <= today
+            and effective_date < last_modified
+        )
+        if effective_stale_date_end_date
+        else False
+    )
+    staleness_12_months_old = (
```
anguskbell, line 98: You've omitted the clause:

> If “Effective stale date due to effective last modified” date is sooner than “Effective stale date due to end date” (if present)

transit_odp/publish/requires_attention.py
anguskbell: It's badly named in that case - if a method begins "add", I'm not expecting it to filter rows out. I would prefer that you either renamed the method or called `get_active_live_revisions` explicitly from here (probably the latter).

transit_odp/organisation/querysets.py
marcusher-ito: Using `relativedelta` here returns a `can't adapt type relativedelta` error. 

If the intention of the staleness here is to flag that it hasn't been updated for a year, being one day out every 4 years and still flagging as stale seems like an acceptable risk and still encouraging updating of data. 

This will need to be consistent with how it's calculated in the file downloads or we might end up with discrepancies - how have you calculated it for your csv?

anguskbell: Well I'm doing it in Python not in the ORM, so I can do:
`datetime.date(d.year+1, d.month, d.day)`

But I doubt that's possible in the ORM.

I'm just flagging it up for the record. If you and Paul have tried to do it and got nowhere, that's fair enough.

AndreaIto: Recently I was facing the same problem 
`relativedelta` doesn't work in the Query, this sentence:
`F("modification_datetime") + relativedelta(years=+1)` 
it's not allow in Django!
The only type allow is `datetime.timedelta`

```
type(timedelta(days=365)) -> datetime.timedelta 
type(relativedelta(years=+1)) -> dateutil.relativedelta.relativedelta
```

anguskbell (dismissed):

anguskbell (approved):

PR #215, opened 2023-02-14
BODS-42 select recorded at time for txc xml file filteration

surabhika:
Hi @anguskbell,
Could you please have a look at this PPC issue and suggest if the change I've done is good enough, please?
Screenshot 2023-02-14 at 15 53 27

### DIFF
--- a/transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
+++ b/transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
@@ -68,6 +68,12 @@ class VehicleJourneyFinder:
             return framed_vehicle_journey_ref.dated_vehicle_journey_ref
         return None
 
+    def get_recorded_at_time(self, vehicle_activity: VehicleActivity) -> Optional[str]:
+        vehicle_recorded_at_time = vehicle_activity.recorded_at_time
+        if vehicle_recorded_at_time is not None:
+            return vehicle_recorded_at_time.date()
+        return None
+
     def check_operator_and_line_present(
         self, activity: VehicleActivity, result: ValidationResult
     ) -> bool:
@@ -445,9 +451,14 @@ class VehicleJourneyFinder:
 
         txc_xml = self.get_corresponding_timetable_xml_files(matching_txc_file_attrs)
 
-        if not self.filter_by_operating_period(
-            activity.recorded_at_time.date(), txc_xml, result
-        ):
+        if (recorded_at_time := self.get_recorded_at_time(activity)) is None:
+            result.add_error(
+                ErrorCategory.GENERAL,
+                "Recorded At Time missing in SIRI-VM VehicleActivity",
+            )
+            return None
+
+        if not self.filter_by_operating_period(recorded_at_time, txc_xml, result):
             return None
 
         if (vehicle_journey_ref := self.get_vehicle_journey_ref(mvj)) is None:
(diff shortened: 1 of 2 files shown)


transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
```diff
@@ -446,7 +446,9 @@ def match_vehicle_activity_to_vehicle_journey(
 
         txc_xml = self.get_corresponding_timetable_xml_files(matching_txc_file_attrs)
 
-        if not self.filter_by_operating_period(activity_date, txc_xml, result):
+        if not self.filter_by_operating_period(
+            activity.recorded_at_time, txc_xml, result
```
anguskbell, line 450: I don't think this will work, as `filter_by_operating_period` is expecting a `date` object, not a `datetime` object. (The type hints in the signature tell you this.)
When this function is called with a `datetime` object, it will compare that to `date` objects, raising an exception I think.
You need to use `activity.recorded_at_time.date()` I think.

anguskbell:
Hi @surabhika, thanks for looking into this. I didn't know it was possible for the `RecordedAtTime` to be the day before.

Your fix addresses the date passed to `filter_by_operating_period` but we also need to fix the date passed to `filter_by_operating_profile` in the same file. The latter function deals with the day of the week used to match the vehicle journey, which is where it's going wrong in the example scenario you've outlined above.

transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
surabhika: My bad. changed it now..

surabhika:
Hi @anguskbell,
I've looked into the `filter_by_operating_profile` function and changed it's input to use recorded at time as well. Thanks for the feedback.

transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
```diff
@@ -446,7 +445,9 @@ def match_vehicle_activity_to_vehicle_journey(
 
         txc_xml = self.get_corresponding_timetable_xml_files(matching_txc_file_attrs)
 
-        if not self.filter_by_operating_period(activity_date, txc_xml, result):
+        if not self.filter_by_operating_period(
+            activity.recorded_at_time.date(), txc_xml, result
+        ):
```
anguskbell, line 450: This looks correct for the happy path. However, if `RecordedAtTime` is missing from the incoming packet (because it's malformed), you'll get an exception here when it accesses the `date` member. All the rest of the PPC code defensively checks that elements XML exist before using their content or accessing sub-elements. Can you please add a defensive check here (and below) - either a check on the value of `recorded_at_time` or a `try...except` block, whichever you prefer.

transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
```diff
@@ -463,7 +464,7 @@ def match_vehicle_activity_to_vehicle_journey(
             return None
 
         if not self.filter_by_operating_profile(
-            activity_date, vehicle_journeys, result
+            activity.recorded_at_time.date(), vehicle_journeys, result
         ):
```
anguskbell, line 468: Same comment as above, re accessing the `date` member.

anguskbell (changes requested):

anguskbell (approved):

PR #231, opened 2023-02-22
BODP-5741: Fix bug where service code exemptions were not being appli…

anguskbell:
…ed in the case of unpublished service codes, i.e. service codes that are in OTC but not in BODS.


marcusher-ito (approved):
Yes this makes sense to me, checking if there's an exemption regardless of whether a service is published in BODS or not.

Just missing version.txt bump

marcusher-ito (approved):

anguskbell:
> Yes this makes sense to me, checking if there's an exemption regardless of whether a service is published in BODS or not.
> 
> Just missing version.txt bump

In general I'd prefer to bump the version later, but since it's a quick review I've bumped it now..

PR #238, opened 2023-02-24
BODP-5724: Fix Safari low-contrast proposed date

marcusher-ito:
Safari (showing unset start date with Safari default of today's date and set end date in the future):

Screenshot 2023-02-24 at 16 30 55

Chrome (showing unset start date with Chrome default of `dd/mm/yyyy` and set end date in the future - behaviour unchanged by this addition):
Screenshot 2023-02-24 at 16 31 24


anguskbell (approved):
Looks plausible. Have you checked other screens (e.g. the API service screens in FBODS) to make sure you haven't made inadvertent changes elsewhere?

anguskbell:
Actually your change is restricted to date-type inputs, so if you've checked the other ones, it's probably all right. Strange that the other ones aren't affected though.

marcusher-ito:
Got muddled up with my local CSS overrides in Safari, it **does** change that behaviour on FBODS Browse Timetables but in the same expected way - if the contrast was an accessibility issue on the Seasonal Services form then it would also be an issue here.
Screenshot 2023-02-27 at 10 55 16
Screenshot 2023-02-27 at 10 55 42

anguskbell:
Ok. Sounds like we're good to go then?

PR #239, opened 2023-02-27
BODP-5635: Updates to the "Download Detailed Export"

AndreaIto:
This export contains service codes for the Operator that have been publish to BODS and that have not been publish to BODS but have been register with the OTC.

In *transit_odp/organisation/csv/service_codes.py* The class `ServiceCodeCSV` inherit from the `CSVBuilder` class the method `to_string` string the data from `get_dataset`.

I compare the `OTCService` data with the `TXCFileAttributes` data to determine which data in BODS require attention:
1. Not publish => 100% require attention
2. Publish - stale stuff => require attention
3. Publish register in OTCservice => Not require attention

In the csv removed the column `Service Status`, the first Column `Require Attention` can be "Yes" or "No" (default).

In *transit_odp/organisation/querysets.py* we `Concat` a CharField: `licence__number` with a IntegerField: `registration_code`, we should `Cast` the Integer into CharField.

In *transit_odp/organisation/tests/csv/test_service_codes.py* fixed some tests functions and rewrote the `test_csv_output` tested all the staleness cases.

In *transit_odp/publish/requires_attention.py* I am using some functions there was a function: `require_attention` that returned a bool I split this function in 2: `require_attention_data` that return a tuple and `require_attention` return a bool. I use the result by `require_attention_data` as a mask to determine which is the staleness status.

In the *transit_odp/publish/views/timetable/requires_attention.py* I replaced the `to_csv` method with the =to_string= from `CSVbuilder`.

@anguskbell @marcusher-ito @ciaranmccormick


transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,229 @@
...
+        if seasonal_service_status
+        else "Not Seasonal"
+        if seasonal_service_start is None
+        else "Out of Season"
+    )
+
+
+def get_seasonal_service_map(organisation_id: int) -> Dict[str, SeasonalService]:
+    """
+    Get a list of dictionaries which includes all the Seasonal Services
```
anguskbell, line 36: It looks like it returns a dictionary, not a list of dictionaries.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,229 @@
...
+        CSVColumn(
+            header="Published Status",
+            accessor=lambda otc_service: "Published"
+            if otc_service.get("dataset_id")
+            else "Unpublished",
+        ),
+        CSVColumn(
+            header="OTC status",
+            accessor=lambda otc_service: "Registered"
+            if otc_service.get("licecnce_number")
```
anguskbell, line 64: Typo: `licence_number`

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,229 @@
...
+            else "Unregistered",
+        ),
+        CSVColumn(
+            header="Scope Status",
+            accessor=lambda otc_service: "Out of Scope"
+            if otc_service.get("service_code") and not otc_service.get("dataset_id")
+            else "In Scope",
+        ),
```
anguskbell, line 72: Scope Status is about whether the service code is exempted in the service code exemption workflow. I don't see how this logic captures that.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,229 @@
...
-        return OTCService.objects.get_all_without_exempted_ones(self._organisation_id)
+    def get_otc_service_data(self, organisation_id: int) -> None:
+        """
+        Compares an organisation's OTC Services dictionaries list with
+        TXCFileAttributes dictionaries list and the SeasonalService list
+        to determine which OTC Services require attention and which
+        doesn't ie. not live in BODS at all, or live but meeting new
+        Staleness conditions.
+        """
+        otc_map = get_otc_map(organisation_id)
```
marcusher-ito, line 188: This function in `transit_odp/publish/requires_attention.py` calls `OTCService.objects.get_otc_data_for_organisation(org_id)` which excludes Exempted and Out of Season services from what's returned, so not sure how using this can give you back all the data you need with those already excluded?

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,229 @@
...
+                    service,
+                    file_attribute,
+                    seasonal_service,
+                    staleness_status,
+                    require_attention,
+                )
+            elif requires_attention(service, file_attribute):
+                require_attention = "Yes"
+                rad = requires_attention_data(service, file_attribute)
+                staleness_status = STALENESS_STATUS[rad.index(True)]
```
anguskbell, line 208: Are you sure that `rad` contains a `True` value? A service code can be not stale and still require attention, so it seems plausible. In this case `index` will raise a `ValueError`.

I think I would write this with a couple of `if` statements instead: more prosaic, but easier to follow.

AndreaIto: I am 100% never hit that exception because 
`rad = requires_attention_data(service, file_attribute)`
is behind an elif `elif requires_attention(service, file_attribute):`
if they are all `0` then this will be `False`

transit_odp/publish/requires_attention.py
```diff
@@ -108,7 +110,16 @@ def requires_attention(service: OTCService, file_attribute: TXCFileAttributes) -
...
-    return any((staleness_otc, staleness_end_date, staleness_12_months_old))
+
+    return (
+        staleness_end_date,
+        staleness_12_months_old,
+        staleness_otc,
+    )
+
+
+def requires_attention(service: OTCService, file_attribute: TXCFileAttributes) -> bool:
```
anguskbell, line 121: This function is about determining staleness, which is just one category of requiring attention, so it would be better renamed to `is_stale`.

transit_odp/publish/requires_attention.py
```diff
@@ -45,7 +45,9 @@ def _update_data(object_list: List[Dict[str, str]], service: OTCService) -> None
     )
 
 
-def requires_attention(service: OTCService, file_attribute: TXCFileAttributes) -> bool:
+def requires_attention_data(
```
anguskbell, line 48: This function is about evaluating the staleness criteria, which is just one aspect of requiring attention, so it would be better renamed to `evaluate_staleness` or similar.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,251 @@
...
+                if exemption or (
+                    seasonal_service and not seasonal_service.seasonal_status
+                ):
+                    require_attention = "No"
+                else:
+                    require_attention = "Yes"
+            elif service and requires_attention(service, file_attribute):
+                rad = requires_attention_data(service, file_attribute)
+                staleness_status = STALENESS_STATUS[rad.index(True)]
+                require_attention = "Yes"
```
anguskbell, line 237: You still need to check for seasonality and for service code exemptions here. If it's exempted for either reason, it doesn't count as requiring attention.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,251 @@
...
+                and seasonal_service.start - timedelta(days=42),
+                "effective_stale_date_end_date": file_attribute
+                and file_attribute.effective_stale_date_end_date,
+                "effective_stale_date_last_modified_date": file_attribute
+                and file_attribute.effective_stale_date_last_modified_date,
+                "last_modified_lt_effective_stale_date_otc": service
+                and (
+                    service.last_modified.date()
+                    < service.effective_stale_date_otc_effective_date
+                ),
```
anguskbell, line 198: This should be comparing the last modified date of the `TXCFileAttribute` object.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,251 @@
...
-    def __init__(self, organisation_id: int):
-        self._organisation_id = organisation_id
-        self.org_service_code_to_dataset = list(
-            TXCFileAttributes.objects.filter(
-                revision__dataset__organisation_id=organisation_id
-            )
-            .get_active_live_revisions()
-            .select_related("revision__dataset")
-            .values_list("service_code", "revision__dataset_id")
+def add_seasonal_service_status(otc_service: dict) -> str:
```
anguskbell, line 21: You're not adding anything here, you're getting the seasonal service status. Should be called `get_seasonal_service_status`.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,251 @@
...
+            header="Scope Status",
+            accessor=lambda otc_service: "Out of Scope"
+            if otc_service.get("scope_status")
+            else "In Scope",
+        ),
+        CSVColumn(
+            header="Seasonal Status",
+            accessor=lambda otc_service: add_seasonal_service_status(otc_service)
+            if otc_service.get("seasonal_status") is not None
+            else None,
```
anguskbell, line 99: The `add_seasonal_service_status` function (which I've requested you rename above!) actually gives you what you want to output here. You shouldn't ever be outputting `None`.

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,251 @@
...
+            accessor=lambda otc_service: otc_service.get("require_attention"),
+        ),
+        CSVColumn(
+            header="Published Status",
+            accessor=lambda otc_service: "Published"
+            if otc_service.get("dataset_id")
+            else "Unpublished",
+        ),
+        CSVColumn(
+            header="OTC status",
```
anguskbell, line 84: I'm sure they would rather this was capitalised, since all the other statuses are.

```suggestion
            header="OTC Status",
```

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -1,82 +1,101 @@
...
 
+    assert len(queryset) == len(expected_pairs_org1)
     queryset_pairs = [
-        (service.licence.number, service.service_code) for service in queryset
+        (service["licence_number"], service["service_code"]) for service in queryset
     ]
     assert sorted(queryset_pairs) == sorted(expected_pairs_org1)
 
 
 def test_missing_service_code():
```
anguskbell, line 96: Not sure why this is called `test_missing_service_code`. In fact I'm not really sure what it's testing.

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -121,92 +138,308 @@ def test_registered_service_code():
...
-    assert row[2] == service_code
-    assert row[3] == service_number
-    assert row[4] in [f"{dataset1.id};{dataset2.id}", f"{dataset2.id};{dataset1.id}"]
+    assert len(queryset) == 1
+    row = queryset[0]
+    assert row["licence_number"] == licence_number
+    assert row["service_code"] == service_code.replace(":", "/")
+    assert row["line_number"] == service_number
+    # only 1 service code, 2 TXCFileAttributes
+    assert row["dataset_id"] == dataset2.id
```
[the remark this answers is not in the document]
AndreaIto: yes correct! the behaviour has changed, in the original code we had this method:

``` 
# transit_odp/organisation/csv/service_codes.py

-    def to_csv(self):
-        string_ = io.StringIO()
-        writer = csv.writer(string_, quoting=csv.QUOTE_ALL)
-        writer.writerow(
-            ("Service Status", "Licence Number", "Service Code", "Line", "Dataset ID")
-        )

-        for otc_service in self.get_queryset():
-            row = self.get_row(otc_service)
-            if row[0] == self.MISSING_DATA:
-                writer.writerow(row)
-
-        for otc_service in self.get_queryset():
-            row = self.get_row(otc_service)
-            if row[0] == self.REGISTERED:
-                writer.writerow(row)
-
-        string_.seek(0)
-        return string_
```
returned the `string_ => io.StringIO`, now to be consistent with the other csvs in BODS, I am using the `CVSBuilder` in this class there is a method `to_string()` return a string  
```
        output = csvfile.getvalue()
        csvfile.close()
        logger.info(prefix + "File successfully closed")
        return output
```
output here is a string

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -121,92 +138,308 @@ def test_registered_service_code():
...
-        )
-    )
+
+    assert csv_output["row1"][0] == '"No"'
+    assert csv_output["row1"][1] == '"Unpublished"'
+    assert csv_output["row1"][2] == '"Registered"'
+    assert csv_output["row1"][3] == '"In Scope"'
+    assert csv_output["row1"][4] == '"Out of Season"'
+    assert csv_output["row1"][5] == '"Not Stale"'
+    assert csv_output["row1"][14] == '""'
```
anguskbell, line 367: It seems arbitrary that you're testing the output of columns 0-5, 14-16 in all cases. In several of those cases, you don't care about columns 14-16 as they're blank.

Can you add some asserts that cover the other columns, where it's appropriate, please - at least the seasonal start and end dates, and "Effective stale date due to end date". Only where it's relevant, not in every row!

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -121,92 +138,308 @@ def test_registered_service_code():
...
+    ServiceModelFactory(
+        licence=otc_lic,
+        registration_number=service_codes[5],
+        service_number=service_numbers[5],
+        effective_date=datetime.datetime(2021, 1, 24),
+    )
+    SeasonalServiceFactory(
+        licence=bods_licence,
+        registration_code=service_codes[5][-1:],
+    )
```
anguskbell, line 276: Bit risky not specifying a start and end date - the factory might randomly generate dates that make today in season!

Applies in a few cases below too.

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -121,92 +138,308 @@ def test_registered_service_code():
...
+    assert csv_output["row4"][5] == '"Stale - OTC Variation"'
+    assert csv_output["row4"][14] == '"2023-06-24"'
+    assert csv_output["row4"][15] == '"False"'
+    assert csv_output["row4"][16] == '"2023-02-10"'
+
+    assert csv_output["row5"][0] == '"Yes"'
+    assert csv_output["row5"][1] == '"Published"'
+    assert csv_output["row5"][2] == '"Registered"'
+    assert csv_output["row5"][3] == '"In Scope"'
+    assert csv_output["row5"][5] == '"Stale - End date passed"'
```
anguskbell, line 405: I can see in a few cases you're not testing the value of Seasonal Status. I think that's because there's a bug in your code that writes it!

In this case, and the cases below, it looks like you've created a seasonal service and so I'd expect a status of Out of Season.

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -121,92 +138,308 @@ def test_registered_service_code():
...
+
+    assert csv_output["row9"][0] == '"No"'
+    assert csv_output["row9"][1] == '"Published"'
+    assert csv_output["row9"][2] == '"Unregistered"'
+    assert csv_output["row9"][3] == '"In Scope"'
+    assert csv_output["row9"][4] == '""'
+    assert csv_output["row9"][5] == '"Not Stale"'
+    assert csv_output["row9"][14] == '"2024-02-24"'
+    assert csv_output["row9"][15] == '""'
+    assert csv_output["row9"][16] == '""'
```
anguskbell, line 445: Can you add a test for "Stale - OTC Variation" please? It's the only one missing!

transit_odp/organisation/csv/service_codes.py
AndreaIto: `seasonal_status` return a `bool` can be `True` or `False` if there is an `SeasonalService` for that `service_code`. This export is about `all` from `OTCService` and `TXCFileAttributes`. If there is no `seasonal_service` this row should be `None`. I think I have to leave as it is

transit_odp/organisation/tests/csv/test_service_codes.py
AndreaIto: Me either!! it's part of the old code! In the previous export there was a column `Service Status`, this test was about `MISSING DATA` case. Today it doesn't test anything at all!! Happy to remove it!

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -90,20 +109,18 @@ def test_missing_service_code():
...
-    assert row[0] == ServiceCodesCSV.MISSING_DATA
-    assert row[1] == licence_number
-    assert row[2] == otc_service_code
-    assert row[3] == otc_service_number
-    assert row[4] == ""
+    row = queryset[0]
+    assert len(queryset) == 2
+    assert row["licence_number"] == licence_number
+    assert row["service_code"] == otc_registration_number
+    assert row["line_number"] == otc_service_number
```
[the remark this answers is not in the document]
AndreaIto: Same as above. In the `old` column `Service Status` can be `MISSING DATA` and `REGISTERED`.
Since this column doesn't exist anymore, it doesn't make sense keep this test case too!

transit_odp/organisation/csv/service_codes.py
```diff
@@ -1,59 +1,253 @@
...
+                "effective_seasonal_start_date": seasonal_service
+                and seasonal_service.start - timedelta(days=42),
+                "effective_stale_date_end_date": file_attribute
+                and file_attribute.effective_stale_date_end_date,
+                "effective_stale_date_last_modified_date": file_attribute
+                and file_attribute.effective_stale_date_last_modified_date,
+                "last_modified_lt_effective_stale_date_otc": service
+                and file_attribute
+                and (
+                    file_attribute.effective_stale_date_last_modified_date
```
anguskbell, line 190: The definition says:

> If last modified date is less than Effective stale date due to OTC effective date
> then TRUE

so you should be using the last modified date on the TXCFileAttributes object, not the "Effective stale date due to effective last modified date".

transit_odp/organisation/tests/csv/test_service_codes.py
```diff
@@ -1,212 +1,640 @@
...
+    start = 0
+    i = 0
+    result.update({"header": result_list[start:CSV_NUMBER_COLUMNS]})
+    start += CSV_NUMBER_COLUMNS
+    for end in range(2 * CSV_NUMBER_COLUMNS, len_result_list + 1, CSV_NUMBER_COLUMNS):
+        result.update({f"row{i}": result_list[start:end]})
+        start = end
+        end += start
+        i += 1
+    return result
```
anguskbell, line 36: This could trivially be implemented using `csv.reader` but I don't want to waste time rewriting it.

For the record, incrementing `end` is pointless as the value is immediately overwritten by the `range` generator.

transit_odp/organisation/tests/csv/test_service_codes.py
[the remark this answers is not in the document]
anguskbell: Actually my question was about the dataset IDs. It used to be a comma-separated list, but now it's not. I don't see anything in the ticket which makes it ok to break that.

I wanted your opinion on how long it would take to fix this regression.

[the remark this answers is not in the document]
AndreaIto: Sorry Angus Are talking about this one?

Screenshot 2023-03-01 at 12 21 43

it's `semicolon-list`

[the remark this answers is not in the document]
anguskbell: So the output *is* a semi-colon-separated list?

The reason I thought it wasn't was because you had a unit test that output only one of the two dataset IDs.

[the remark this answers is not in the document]
AndreaIto: Actually I am asking to you, Is this `semicolon-separated list` which are you talking about?

In my test I am testing the `only one` `dataset ID` in BODS that come from `revision`.
How many dataset IDs is suppose to test for you?

anguskbell (approved):
I'm approving the changes so far. The remaining issue about duplicate service codes in multiple datasets is TBD and can be done as a separate commit.

PR #248, opened 2023-03-03
story/BODS-230 + BODS-265: Recommended actions + copy for when fares file is uploaded and not compliant

ndemenezes:


transit_odp/fares/views/update.py
```diff
@@ -99,18 +100,19 @@ def get(self, request, *args, **kwargs):
...
+        #     return redirect(
+        #         "fares:feed-draft-exists",
+        #         pk=self.object.dataset_id,
+        #         pk1=self.object.dataset.organisation_id,
+        #     )
+        # else:
+        # Ensure GET returns the preview step
+        self.object.dataset.start_revision()
+        self.storage.reset()
+        return self.render(self.get_form())
```
surabhika, line 115: I think there is no harm in keeping the feed draft exists view if a user tries to create a new draft when there is already a draft existing.
Although, This will not be reproduced as the previous draft is been deleted.

ndemenezes: @surabhika Have tested and pushed the code back in.

surabhika: Not that it matters much, but the return should be inside else block

transit_odp/fares/tasks.py
```diff
@@ -62,7 +62,7 @@ def task_run_fares_pipeline(self, revision_id: int, do_publish: bool = False):
 
         task_download_fares_file(task.id)
         task_run_antivirus_check(task.id)
-        task_run_fares_validation(task.id)
+        # task_run_fares_validation(task.id)
```
surabhika, line 65: This needs be uncommented

transit_odp/fares/tasks.py
```diff
@@ -172,6 +174,27 @@ def task_run_fares_validation(task_id):
...
+    org_id_list = Dataset.objects.filter(id=revision.dataset_id).values_list(
+        "organisation_id", flat=True
+    )
+
+    adapter.info("Setting validation errors.")
+    fares_validator_obj = FaresXmlValidator(file_, org_id_list[0], revision.id)
+    fares_validator_obj.set_errors()
+
+    task.update_progress(50)
+
```
surabhika, line 196: Please add some logs saying that fares validation started and ended.

transit_odp/fares_validator/views/validate.py
```diff
@@ -48,6 +52,10 @@ def set_errors(self):
             response = JsonResponse(
                 serializer.data, safe=False, status=status.HTTP_201_CREATED
             )
+        # For 'Update data' flow
+        FaresValidationResult.objects.filter(
+            revision_id=self.pk2, organisation_id=self.pk1
+        ).delete()
         FaresValidationResult.create_validation_result(
             revision_id=self.pk2, org_id=self.pk1, violations=violations
         ).save()
```
surabhika, line 61: Please add `with transaction.atomic():` for all or nothing for this entire deletion and addition of records for both the tables i.e from line 41 to 61..

transit_odp/fares/tasks.py
ndemenezes: Done this change

transit_odp/fares_validator/views/validate.py
ndemenezes: Done this change

surabhika (approved):

PR #269, opened 2023-03-20
Story/bods 565 Unexpected stale reason 'Stale - OTC Variation' 

surabhika:


transit_odp/timetables/tests/test_csv.py
```diff
@@ -8,7 +8,6 @@
     DataQualityReportFactory,
     PTIValidationResultFactory,
 )
-from transit_odp.organisation.csv import EmptyDataFrame
```
ndemenezes, line 11: These two test cases: test_empty_otc_services and test_empty_txc_services fail because this line is removed.

surabhika: Didn't mean to remove this! Added it back

PR #298, opened 2023-04-06
Story/bods 351 Update PPC logic to allow for duplicate Serviced Organisation journey code

surabhika:


transit_odp/timetables/transxchange.py
```diff
@@ -315,6 +315,13 @@ def get_vehicle_journeys(self):
         xpath = ["VehicleJourneys", "VehicleJourney"]
         return self._root.get_elements(xpath)
 
+    def get_serviced_organisations(self):
+        """
+        Get all Serviced Organisations
+        """
+        xpath = ["ServicedOrganisations", "ServicedOrganisation"]
+        return self.__root.get_element(xpath)
```
ndemenezes, line 323: **test_filter_by_days_of_operation** test case fails with this line => Tried it like this and it works:
`return self._root.get_elements(xpath)`

ndemenezes (changes requested):

transit_odp/timetables/transxchange.py
surabhika: ohh yes that's a typo I thought I fixed. thanks

surabhika: That's fixed now!

ndemenezes (approved):

pritish-ranjan01 (approved):

PR #302, opened 2023-04-11
Release 1.25.0 into dev

pritish-ranjan01:


version.txt
```diff
@@ -1 +1 @@
-v1.24.8
+v1.24.9
```
surabhika, line 1: Since its the next release now, let's start from v1.25.0.

PR #350, opened 2023-05-30
Story/BODS-310 Default browse screens to active datasets

surabhika:


ndemenezes (commented):
AC6 is not working for me - 'Published' status is not showing up as a default:

image

surabhika:
@ndemenezes your comment should be addressed. Thank you for pointing that out.

PR #355, opened 2023-06-01
story/BODS-955: Remove inactive datasets from Fares data catalogue

ndemenezes:


transit_odp/fares/querysets.py
```diff
@@ -70,6 +71,9 @@ def get_active_fares_files(self):
             .add_string_user_type()
             .add_organisation_name()
             .add_compliance_status()
+            .exclude(
+                fares_metadata_id__revision__dataset__live_revision__status=INACTIVE
```
surabhika, line 75: You can just do `fares_metadata_id__revision__status=INACTIVE`

surabhika (approved):

PR #363, opened 2023-06-06
story/BODS-314: Updates to FBODS - Browse Data - location data

ndemenezes:


surabhika:
AC2 in the story is not really a part of this change, right?

ndemenezes:
> 

In forms.py and filters.py, I've removed 'No vehicle activity' from the status filter.

surabhika (approved):

PR #365, opened 2023-06-08
Fix for BODS-994 and BODS-1004

surabhika:


transit_odp/browse/tests/test_fares.py
ndemenezes, line 1: @surabhika On line 6 in this file, is it supposed to be 'TestBaseAVLSearchView' instead?

ndemenezes (changes requested):

transit_odp/browse/tests/test_fares.py
surabhika: Yes! It should be. Changing that now. Thanks!

ndemenezes (approved):

pritish-ranjan01 (approved):

PR #366, opened 2023-06-09
story/BODS-383: considering error status data feeds as published

pritish-ranjan01:


surabhika:
Please resolve the conflicts and fix test cases as they are failing because of the change in queryset.

transit_odp/organisation/querysets.py
```diff
@@ -346,13 +345,9 @@ def add_pretty_status(self):
...
             status=Case(
                 When(
-                    Q(live_revision__status="live"),
+                    Q(live_revision__status__in=["live", "error"]),
                     then=Value("published", output_field=CharField()),
                 ),
-                When(
-                    Q(live_revision__status="error", dataset_type=AVLType),
-                    then=Value(NO_ACTIVITY, output_field=CharField()),
-                ),
```
surabhika, line 355: Now looking at this again, the second when condition should be kept because it's only for AVL that we cant the status "error" to be shown as "published", and not for timetables/fares

surabhika (approved):

PR #375, opened 2023-06-16
story/BODS-383: Handle 'error' status for AVL feeds

pritish-ranjan01:


transit_odp/organisation/querysets.py
```diff
@@ -341,6 +341,19 @@ def add_live_data(self):
...
+                When(
+                    Q(live_revision__status="error"),
+                    then=Value("live", output_field=CharField()),
+                ),
+                default=F("live_revision__status"),
+                output_field=CharField(),
+            )
+        )
+
     def add_pretty_status(self):
```
surabhika, line 357: Line 353 will need to be reverted, right?

pritish-ranjan01: I think, we need the output field for the default..

surabhika (approved):

PR #376, opened 2023-06-19
story/BODS-992: Remove leading/trailing spaces within LTA names

ndemenezes:


transit_odp/browse/views/local_authority.py
```diff
@@ -117,7 +117,7 @@ def object_list_lta_mapping(self, object_list):
         new_object_list = []
         for lta_object in object_list:
             for otc_name, value_name in LTAS_DICT.items():
-                if lta_object.name == otc_name:
+                if lta_object.name.strip() == otc_name:
```
surabhika, line 120: The OTC name from the dictionary(which will be changed to a model later) should also to be trimmed, because it could also have extra white spaces as it will be entered by a user in Django admin.

surabhika: On second thought, along with this the queryset also needs to be changed. If the name field from the database has extra spaces then the comparison on line 127 would not give you that LTA. And same will happen if there is a space in the LTA_DICT name.

ndemenezes: @surabhika Completed and pushed the changes.

PR #381, opened 2023-06-22
story/BODS-988

pritish-ranjan01:


transit_odp/otc/admin.py
```diff
@@ -81,3 +82,30 @@ def has_add_permission(self, request, instance=None):
...
+        model = LocalAuthority
+        fields = ["name", "ui_lta_name", "atco_code"]
+
+
+@admin.register(LocalAuthority)
+class LocalAuthorityAdmin(admin.ModelAdmin):
+    list_per_page = 50
+    list_display = ("name", "ui_lta_name", "atco_code")
+    search_fields = ("name", "ui_lta_name")
+    list_filter = ("name",)
```
surabhika, line 98: I think the list displayed on Django admin should be ordered by name. ATM it shows the list as it's saved in the database.

surabhika (changes requested):

transit_odp/otc/admin.py
pritish-ranjan01: Done

surabhika (approved):

PR #391, opened 2023-07-11
Story/BODS-1033 Relax rules to allow where more than one file belongs to a combination of NOC and line and journey code

surabhika:


ndemenezes (changes requested):
@surabhika The text on the guidance page hasn't been changed yet for Step 5 (AC2 on [BODS-1033)](https://kpmgengineering.atlassian.net/browse/BODS-1033).


EDIT: that should be done now @ndemenezes

transit_odp/avl/post_publishing_checks/daily/vehicle_journey_finder.py
```diff
@@ -622,6 +654,10 @@ def match_vehicle_activity_to_vehicle_journey(
             ):
                 return None
 
+        if len(vehicle_journeys) > 1:
```
pritish-ranjan01, line 657: Can we combine this with the check just before this using an "or" operator?

ndemenezes (approved):

PR #394, opened 2023-07-12
Alter task update to include tags

carmstrong23:
Tags are required in the infrastructure for monitoring and cost attribution. The aws actions strip out the tags so altering to use a describe including tags and jq to modify the image based on container name


RukayyahP (dismissed):

carmstrong23:
@qaraaslam @pritish-ranjan01  - updated PR to remove static reference to squid!

qaraaslam (approved):

PR #399, opened 2023-07-17
story/BODS-1050

pritish-ranjan01:


transit_odp/otc/models.py
```diff
@@ -116,3 +119,9 @@ def __str__(self) -> str:
     class Meta:
         verbose_name_plural = "Local Authorities"
         verbose_name = "Local Authority"
+
+
+class InactiveService(models.Model):
+    registration_number = models.CharField(max_length=20, blank=False, null=False)
```
surabhika, line 125: registration number should be a unique field in the model as a registration can be picked up multiple times of the effective date is in the range for last 3 days.

transit_odp/otc/client/otc_client.py
```diff
@@ -118,6 +118,29 @@ def get_latest_variations_since(self, when: datetime) -> List[Registration]:
...
+
+            for page in range(2, response.page.total_pages + 1):
+                response = self._make_request(
+                    page=page,
+                    regNo=registration_code,
+                    latestVariation=True,
+                )
+                variations += response.bus_search
+
+            return variations
```
surabhika, line 142: The return statement should be outside the for loop. otherwise this function will exit after the first item in the loop.

transit_odp/otc/models.py
surabhika: And one more reason it should be unique is that we should only have one row per registration number even if there are more than one variations returned as a response from the OTC API.

surabhika (changes requested):

transit_odp/otc/registry.py
```diff
@@ -184,6 +196,12 @@ def update_registered_variations(self, variation: Registration) -> None:
             if variation.effective_date:
                 if variation.effective_date <= date.today():
                     self.update(variation)
+                else:
+                    InactiveService.objects.create(
+                        registration_number=variation.registration_number,
+                        registration_status=variation.registration_status,
+                        effective_date=variation.effective_date,
+                    )
```
surabhika, line 204: This will fail when trying to insert same registration again as its now unique.

surabhika (changes requested):

surabhika (approved):

PR #414, opened 2023-07-31
fix/BODS-1225: AVL: Data feed details page has a line that is shorter

ndemenezes:


surabhika (approved):

surabhika:
Do not merge until v1.28 is in production.

PR #422, opened 2023-08-02
story/BODS-1310: Enabling profiling/tracing using ddtrace

pritish-ranjan01:
* Includes updated version of attrs python library
* New ddtrace library
* Updated manage.py and celery.py to enable base tracing/profile


transit_odp/taskapp/celery.py
PCridKPMG, line 1: You'll need to move lines 8 (from ddtrace import patch_all) and 10 (patch_all()) to before you import celery, so move'em to line's 3 and 5

From https://ddtrace.readthedocs.io/en/stable/basic_usage.html :

Note: To ensure that the supported libraries are instrumented properly in the application, they must be patched prior to being imported. So make sure to call patch_all before importing libraries that are to be instrumented.

PCridKPMG (commented):
One proposed change in cerlery.py to move import and patch_all further up the script _before_ celery and django imports - see note

transit_odp/taskapp/celery.py
pritish-ranjan01: Done

PCridKPMG (approved):

surabhika (approved):

PR #433, opened 2023-08-11
story/BODS-788: SPIKE: NOT TO BE DEPLOYED IN TEST AND HIGHER ENVS

pritish-ranjan01:


surabhika:
Think you should also do the changes for a single xml file so that the publish flow does not break.

PR #434, opened 2023-08-11
story/BODS-1117: Add Organisation name for unpublished entries on the LTA service code report

ndemenezes:


transit_odp/otc/querysets.py
```diff
@@ -381,6 +381,26 @@ class OperatorQuerySet(QuerySet):
     def add_service_count(self):
         return self.annotate(service_count=Count("services"))
 
+    def get_otc_organisation_name(self, licence_number):
+        from transit_odp.otc.models import Licence as OTCLicence
```
pritish-ranjan01, line 385: Is there a specific reason for local imports?

transit_odp/otc/querysets.py
```diff
@@ -381,6 +381,26 @@ class OperatorQuerySet(QuerySet):
...
+                number=licence_number
+            ).values_list("id", flat=True)
+            otc_operator_ids = OTCService.objects.filter(
+                licence_id__in=licence_subquery
+            ).values_list("operator_id", flat=True)
+
+            return (
+                self.filter(operator_id__in=otc_operator_ids)
+                .values_list("operator_name", flat=True)
+                .first()
```
pritish-ranjan01, line 399: Does using .first() give the desired result everytime?

transit_odp/otc/querysets.py
ndemenezes: @pritish-ranjan01 These two models are only used within this function that's why I decided to put them here. Is it better to put them at the top of the file instead?

pritish-ranjan01: Function level imports are usually discouraged because this could lead to repeated imports when the function is called multiple times. So, better to have it imported at top level.

transit_odp/otc/querysets.py
ndemenezes: @pritish-ranjan01 It should, as the id is a pk so there should only be one operator name returned.

transit_odp/browse/views/local_authority.py
```diff
@@ -32,6 +33,21 @@ def get_seasonal_service_status(otc_service: dict) -> str:
     return "In Season" if seasonal_service_status else "Out of Season"
 
 
+def get_operator_name(otc_service: dict) -> str:
+    otc_licence_number = otc_service.get("otc_licence_number")
+    otc_service["operator_name"] = OTCOperator.objects.get_otc_organisation_name(
```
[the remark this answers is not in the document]
ndemenezes: @pritish-ranjan01 Done

pritish-ranjan01 (approved):

ndemenezes:
@pritish-ranjan01 @surabhika This is now ready to be reviewed.

surabhika (approved):

PR #454, opened 2023-08-23
fix/BODS-1136: Link within PTI validation notification is not accessIble when not logged in

ndemenezes:


surabhika:
The link to dataset will still not be accessible when not logged in. This change will only make the report link accessible. If that's the only change we want, I am happy with this.

ndemenezes:
> The link to dataset will still not be accessible when not logged in. This change will only make the report link accessible. If that's the only change we want, I am happy with this.

@surabhika The validation report link should be accessible, whereas the link to the dataset should be only accessible to the logged in operator.

surabhika (approved):

PR #526, opened 2023-10-11
fix/BODS-1591: Fix flexible service bug issue

ndemenezes:


transit_odp/data_quality/pti/functions.py
```diff
@@ -299,6 +300,8 @@ def has_flexible_service_classification(context, services):
                         return False
             else:
                 return False
+        else:
```
pritish-ranjan01, line 303: Can this bit be simplified without these many nested conditions?
Please see below.

if not flexible_journey_pattern_list and not booking_arrangements_list:
    return True

service_classification_list = service.xpath("x:ServiceClassification", namespaces=ns)
if not service_classification_list:
    return False

for service_classification in service_classification_list:
    if service_classification.xpath("x:Flexible", namespaces=ns):
        return True

return False

ndemenezes: Done.

pritish-ranjan01 (approved):

PR #558, opened 2023-10-30
round the data quality score to lower number instead of higher rounded number

alvak1:
Testing instructions

- Publish a timetable and view the data quality score in the review page
- Access below page and verify the data quality score
  Publish endpoint -> Review Published Dataset -> Select a published timetable


PCridKPMG:
@alvak1 Sorry sir, don't think this is the solution. Your problem is I believe in transit_odp/common/templatetags/filters.py in

`
@register.filter("percentage")
def percentage(text, arg=-1):
    return floatformat(text * 100.0, arg=arg) + "%"
`

On the offending pages this is sent either 99.9,"0" or 99.9,0 and its there that the rounding is happening. Pretty sure that if arg is 0 or "0" and the value is 99 < x < 100 we want to round down to 99 or if the value is 0 < x < 1 we want to round up to 1

alvak1:
Closing the PR as another approach is required to fix the problem

PR #586, opened 2023-11-13
Changes to display data quality score with expected percentage values

alvak1:
Testing instructions

- Publish a timetable and view the data quality score in the review page
- Access below page and verify the data quality score
   Publish endpoint -> Review Published Dataset -> Select a published timetable

### DIFF
--- a/transit_odp/common/templatetags/filters.py
+++ b/transit_odp/common/templatetags/filters.py
@@ -18,10 +18,11 @@ def percentage(text, arg=-1):
 
     # If value is 0.1 then 1% should be returned
     # If value is 99.8 then 99% should be returned
-    if 0 < percentage_value < 1:
-        return "1%"
-    elif 99 < percentage_value < 100:
-        return "99%"
+    if int(arg == 0):
+        if 0 < percentage_value < 1:
+            return "1%"
+        elif 99 < percentage_value < 100:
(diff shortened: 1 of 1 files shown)


transit_odp/common/templatetags/filters.py
```diff
@@ -8,7 +8,21 @@
...
+        score = float(text)
+    except (TypeError, ValueError):
+        # Handle the case where the conversion to float fails (e.g., if value is not numeric)
+        return text
+
+    percentage_value = score * 100.0
+
+    # If value is 0.1 then 1% should be returned
+    # If value is 99.8 then 99% should be returned
+    if 0 < percentage_value < 1:
```
PCridKPMG, line 21: @alvak1 You should only return these if int(arg==0), i.e.

if 0 < percentage_value < 1 and int(arg == 0):

if elif 99 < percentage_value < 100 and int(arg == 0):

PCridKPMG (commented):
Needs a change so it only does the logic if int(arg)==0

transit_odp/common/templatetags/filters.py
alvak1: @PCridKPMG  done

PCridKPMG (approved):

PR #596, opened 2023-11-15
fix to display correct value matching the report status

alvak1:
The fix should not round values above 0.995 to 1 there by returning 99.5% instead of 100%


PCridKPMG (changes requested):
I see what you're getting at here but its going to give a rounded down value for - all numbers. I think you need to use similar logic to the other ticket so it rounds to 1dp for everything between 0.1 and 99.9 and the rounds everything greater than zero and less than 0.1% to 0.1% and everything gerater than 99.9 and less than 100 to 99.9%

PCridKPMG (approved):

PR #611, opened 2023-11-20
removing migration script that updates service type column values

alvak1:
removing migration script that updates service type column values


pritish-ranjan01 (approved):

PCridKPMG:
@alvak1 @pritish-ranjan01 any idea what theimplications of pushing this to dev when dev has already had the migration applied will be?

PR #612, opened 2023-11-21
fix to display search items for autocomplete

alvak1:
fix to display search items for autocomplete


amitkpmguk (approved):
Looks good to me 👍

PR #627, opened 2023-11-28
Story/bods 1620

amitkpmguk:
PR includes code for bug of local authorities list page where on authorities page number of authorities were inconsistent.

### DIFF
--- a/transit_odp/browse/views/local_authority.py
+++ b/transit_odp/browse/views/local_authority.py
@@ -1,7 +1,8 @@
 from datetime import timedelta
 from typing import Dict, Optional
 
-from django.db.models import Subquery
+from django.db.models import Subquery, Count
+
 from django.db.models.functions import Trim
 from django.http import HttpResponse
 from django.views import View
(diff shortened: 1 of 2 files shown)


transit_odp/browse/views/local_authority.py
```diff
@@ -142,49 +143,41 @@ def get_context_data(self, **kwargs):
         ids_list = {}
```
alvak1, line 143: This variable can be removed

transit_odp/browse/views/local_authority.py
```diff
@@ -142,49 +143,41 @@ def get_context_data(self, **kwargs):
...
-
-        for lta_id_list in ids_list.values():
-            for lta in all_ltas_current_page:
-                if lta.id in lta_id_list:
-                    setattr(lta, "auth_ids", lta_id_list)
+            lta_list = (
+                self.model.objects.annotate(ui_lta_name_trimmed=Trim("ui_lta_name"))
+                .filter(ui_lta_name_trimmed=lta.ui_lta_name_trimmed)
+                .all()
+            )
```
alvak1, line 150: Can we run the query once to obtain all the ltas outside the loop instead of invoking a db call within the for loop? We could re-use `all_ltas` but instead of using the queryset we use the model objects.

transit_odp/browse/views/local_authority.py
amitkpmguk: Resolved

transit_odp/browse/views/local_authority.py
```diff
@@ -139,77 +140,70 @@ def get_context_data(self, **kwargs):
...
-                    )
 
         names = []
         name_set = set()
+        ui_lta_wise_ltas = {}
 
-        all_ltas = self.get_queryset()
+        all_ltas = self.get_model_objects().all()
         for lta in all_ltas:
+            ui_lta_wise_ltas.setdefault(lta.ui_lta_name_trimmed, []).append(lta)
```
alvak1, line 150: As discussed - could we verify if there are automated test cases to check the paginated record count when that page has multiple local authorities with the same UI name.

transit_odp/browse/views/local_authority.py
```diff
@@ -139,77 +140,70 @@ def get_context_data(self, **kwargs):
...
-                        context["services_require_attention_percentage"] = 0
-                    setattr(
-                        lta,
-                        "services_require_attention_percentage",
-                        context["services_require_attention_percentage"],
-                    )
 
         names = []
         name_set = set()
+        ui_lta_wise_ltas = {}
```
alvak1, line 146: Rename this variable to something along `lta_list_per_ui_ltas`

alvak1 (approved):
Approved

PR #628, opened 2023-11-28
Line names can be multiple so passed these through loop

amitkpmguk:
PR contains fix related to line names not visible when multiple line names are present in service code


alvak1 (dismissed):
Approved

brahmaduttau (approved):

PR #645, opened 2023-12-11
changes to create ui ltas and map it to local authorities

alvak1:
changes to create ui ltas and map it to local authorities


ankitmahajan1 (approved):
looks good

PR #782, opened 2024-02-12
story/BODS-2280

pritish-ranjan01:
Handled issue when multiple serviced org refs are present in same WorkingDays element
Added test cases


alvak1 (approved):
Approved

PR #830, opened 2024-03-05
fix/BODS-165 + fix/BODS-129 + fix/BODS-148: Error not underlined as per GDS

ndemenezes:
Same fix for BODS-165 + BODS-129 + BODS-148


amitkpmguk (approved):
PR looks good to me 👍

PR #833, opened 2024-03-06
fix for an error when timetable visualiser flag is off

alvak1:
fix for an error when timetable visualiser flag is off


ankitmahajan1 (approved):
approved

pritish-ranjan01 (approved):

PR #861, opened 2024-03-15
[BODS-3532] updated in-scope out-of-scope logic for lta report

amitkpmguk:


transit_odp/browse/common.py
```diff
@@ -1,18 +1,22 @@
...
+import pandas as pd
+
 from transit_odp.naptan.models import AdminArea
-from transit_odp.otc.models import Service
 from transit_odp.organisation.constants import TravelineRegions
-from typing import Dict
+from transit_odp.otc.models import Service
 
 
 def get_all_naptan_atco_map() -> Dict[str, str]:
```
ankitmahajan1, line 10: change the return type for this function 
and we can rename the function as it not returning the map or dict

transit_odp/browse/common.py
```diff
@@ -73,6 +77,9 @@ def get_service_traveline_regions(ui_ltas):
     return "|".join(
         [
             str(TravelineRegions(admin_area.traveline_region_id).label)
-            for admin_area in AdminArea.objects.filter(ui_lta__in=ui_ltas).all()
+            for admin_area in AdminArea.objects.filter(ui_lta__in=ui_ltas)
+            .order_by("traveline_region_id")
```
ankitmahajan1, line 81: we can apply order_by after distinct so less records need to be operated on

amitkpmguk: Updated

transit_odp/browse/common.py
amitkpmguk: Updated

ankitmahajan1 (approved):
approved

PR #869, opened 2024-03-18
Updated logic for the services when they have other UI Lta in scope b…

amitkpmguk:
…ut not the current one, Such services are considered in scope


ankitmahajan1 (approved):
approved

PR #890, opened 2024-03-25
fix to process serviced org code within Holidays element

alvak1:
Process files which contain serviced organisation code within Holidays tag

https://kpmgengineering.atlassian.net/browse/BODS-3667


transit_odp/timetables/dataframes.py
```diff
@@ -746,12 +746,23 @@ def populate_operating_profiles(
         )
         if days_of_operation:
             operational = True
-            serviced_orgs_working_days = days_of_operation.get_element("WorkingDays")
+            serviced_orgs_working_days = days_of_operation.get_element_or_none(
+                "WorkingDays"
+            )
+            if not serviced_orgs_working_days:
+                serviced_orgs_working_days = days_of_operation.get_element("Holidays")
+                if serviced_orgs_working_days:
```
NitishAroraKPMG, line 754: do we need to put same condition twice ... at line 752 also

if not serviced_orgs_working_days

transit_odp/timetables/dataframes.py
```diff
@@ -746,12 +746,23 @@ def populate_operating_profiles(
...
+                    operational = False
```
NitishAroraKPMG, line 755: are we using the flag, anywhere. if yes, resolve the comment.

NitishAroraKPMG (approved):

PR #897, opened 2024-03-27
Update version.txt

brahmaduttau:
- Timetable Catalogue- "Requires Attention" and "Scope Status" Column logic update
- Timetable catalogue showing Scottish service codes as in scope
- Timetable data catalogue - WECA registrations + UI LTA/ATCO relationship
- Null services showing as in scope in timetable data catalogue
- Timetable data catalogue contains duplicate service codes


ankitmahajan1 (approved):
approved

PR #981, opened 2024-04-24
Moving the departure day shift to last in timetable

NitishAroraKPMG:


transit_odp/timetables/utils.py
```diff
@@ -244,10 +244,9 @@ def get_vehicle_journey_codes_sorted(
     """
     Get the vehicle journey codes sorted based on the departure time
     """
-
-    df_vehicle_journey_sorted = df_vehicle_journey_operating.groupby(
-        "vehicle_journey_code"
-    ).apply(lambda x: x.sort_values(by="departure_time"))
+    df_vehicle_journey_sorted = df_vehicle_journey_operating.sort_values(
+        by=["departure_day_shift", "departure_time"], ascending=[True, True]
```
alvak1, line 248: When departure_day_shift exist then the journey should be at the end of the timetable

transit_odp/timetables/transformers.py
```diff
@@ -161,9 +160,12 @@ def transform(self) -> TransformedData:
             ) = transform_service_pattern_to_service_links(  # noqa: E501
                 service_patterns, route_to_route_links, route_links
             )
-            service_patterns = service_patterns.drop_duplicates(
+            service_patterns = service_patterns.reset_index().drop_duplicates(
```
alvak1, line 163: Multiple entries with different service pattern id for the same vehicle journey seems to be stored in db. Also most of these journeys are missing

image

transit_odp/timetables/transformers.py
```diff
@@ -161,9 +160,12 @@ def transform(self) -> TransformedData:
...
                 drop_duplicates_columns_sp
             )
+            service_patterns = service_patterns.set_index(
```
alvak1, line 166: Method `transform_stop_sequence` can be updated to ensure the geometries for routes are calculated only once. `route_hash` can be used to map geometry of unique row and then map these to the relevant service pattern row

transit_odp/transmodel/models.py
```diff
@@ -78,7 +78,6 @@ class ServicePattern(models.Model):
 
     class Meta:
         ordering = ("revision", "service_pattern_id", "line_name")
-        unique_together = ("revision", "service_pattern_id", "line_name")
```
alvak1, line 81: Three test cases are failing

alvak1 (approved):
Approved

PR #1042, opened 2024-05-02
[BODS-4647] Added the registration with status in to_change but variation number …

amitkpmguk:
…zero so that can be deleted from the database

PR contains following bug
https://kpmgengineering.atlassian.net/browse/BODS-4647


ankitmahajan1 (approved):
approved

PR #1047, opened 2024-05-03
Rounding of the seconds and changes in serviced org

alvak1:
* Adding the rounding of the seconds

* Adding the test cases


NitishAroraKPMG (approved):
approved

PR #1048, opened 2024-05-03
Adding the serviced org working days operating and rounding down the minute

NitishAroraKPMG:


NitishAroraKPMG:
Not needed and will be raised in the hotfix branch

PR #1065, opened 2024-05-08
fix/failing-unit-test: Fix line description failing unit test case

ndemenezes:
Fixed the unit test that keeps on failing: `test_line_description_false`


ankitmahajan1 (approved):
approved

PR #1219, opened 2024-06-25
story: Update swagger for CAVL

alainajaffer:


nbetts (commented):
I think the changes look good, just same questions from the other PRs:
1. Where does the CAVL_API_BASE_URL get set if its value is blank in the .env files?
2. Do the changes need to be behind a feature flag in case bods has a release before the new AVL service goes live?

alainajaffer:
> 2\. Do the changes need to be behind a feature flag in case bods has a release before the new AVL service goes live?

Do the changes need to be behind a feature flag in case bods has a release before the new AVL service goes live? @DeannaSharma @ljones92 

Where does the CAVL_API_BASE_URL get set if its value is blank in the .env files? You have to hardcode it in .env to https://avl.test.integrated-data.dft-create-data.com

nbetts (approved):

PR #1225, opened 2024-06-26
story: UI update for AVL publisher flow to include new static IP addresses

alainajaffer:


config/.env.local.template
```diff
@@ -66,6 +66,7 @@ CAVL_CONSUMER_URL=__change_me__
 # Central AVL Service
 # ------------------------------------------------------------------------------
 CAVL_API_BASE_URL=__change_me__
+CAVL_IP_ADDRESS_LIST=
```
nbetts, line 69: where are the IP addresses set if this is empty (and the other config templates)?

alainajaffer: prod ones haven't been determined yet (correct me if I am wrong @DeannaSharma and @ljones92  ) so I assume this value will be added in the .env value hardcoded and provided based on the environment running.

nbetts (approved):

config/.env.local.template
DeannaSharma: Yeah prod IPs haven't been determined. 

I see what Nathan is getting at, if whatever reason these environment variable is not set what behaviour do we want to see? 

Maybe we need a check in `forms.py` where if `ALLOW_LIST_IP_ADDRESSES` is undefined then we do not show the bit of text on the front end about white listing IPs.

alainajaffer: done

PR #1266, opened 2024-07-05
[BODS-4392] updated return type to filtered values

amitkpmguk:
PR includes following story
https://kpmgengineering.atlassian.net/browse/BODS-4392


ankitmahajan1 (approved):
approved

PR #1311, opened 2024-07-18
[BODS-6057] First stop always set as a timing point in the is_timing_point column

ankitmahajan1:


amitkpmguk (approved):
Looks good to me 👍

pritish-ranjan01 (approved):

PR #1330, opened 2024-07-24
[BODS-6129] Password Policy

ankitmahajan1:
https://kpmgengineering.atlassian.net/jira/software/c/projects/BODS/boards/1975?selectedIssue=BODS-6129

### DIFF
--- a/transit_odp/common/password_validation.py
+++ b/transit_odp/common/password_validation.py
@@ -8,19 +8,11 @@ class CustomPasswordValidator:
     Custom password validator for Django authentication system.
 
     This validator checks if a password meets the following criteria:
-    - Minimum length of 8 characters.
     - Contains at least one uppercase letter.
     - Contains at least one lowercase letter.
     - Contains at least one digit.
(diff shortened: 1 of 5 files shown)


transit_odp/common/password_validation.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+        - password (str): The password to validate.
+        - user (object, optional): The user object (default: None).
+
+        Raises:
+        - ValidationError: If the password does not meet any of the validation criteria.
+        """
+
+        # Minimum password length
+        min_length = 8
```
brahmaduttau, line 31: Add this in settings

brahmaduttau (dismissed):

transit_odp/common/password_validation.py
ankitmahajan1: minimum length validation has been removed from this custom validator 
as it is already added in base.py

brahmaduttau (approved):

PR #1385, opened 2024-07-30
BODS-6234 AVL backend switch for get/delete feeds

nbetts:


transit_odp/notifications/client/base.py
```diff
@@ -729,7 +729,9 @@ def send_avl_feed_subscriber_notification(
             f"<id={dataset_id} >"
         )
         feed_status = (
-            "No vehicle activity" if dataset_status == AVL_FEED_DOWN else "Published"
+            "No vehicle activity"
+            if dataset_status == AVLFeedStatus.error.value
```
DeannaSharma, line 733: Could we have a call about this bit of logic - I forsee a little issue but I think it just means a small code change.

I think `dataset_status` comes from the one of the BODS table and reads a column called `avl_feed_status` - I think we should update it to read from the `status` column as that column matches our enum. 

My worry is that when we make this code change any older entries will keep the old feed status enums and would mean this notifications code won't work as expected.

nbetts: Ahh yes, I see what you mean. I'll update it and happy to discuss on a call as well

PR #1442, opened 2024-08-13
[BODS-4379] changes for EP api

ankitmahajan1:


config/settings/base.py
```diff
@@ -601,13 +601,13 @@
 # -------------------------------------------------------------------------------
 EP_API_URL = env(
     "EP_API_URL",
-    default="https://djzdv19waj3wh.cloudfront.net/api/v1/search",
+    default="EP_API_URL",
```
NitishAroraKPMG, line 604: put some dummy value, in order to understand the issue, if it comes
like https://ep_api.url

config/settings/base.py
```diff
@@ -601,13 +601,13 @@
...
 )
 EP_CLIENT_ID = env("EP_CLIENT_ID", default="EP_CLIENT_ID")
 EP_CLIENT_SECRET = env("EP_CLIENT_SECRET", default="EP_CLIENT_SECRET")
 EP_AUTH_URL = env(
     "EP_AUTH_URL",
-    default="https://dev-pdbrd.auth.eu-west-2.amazoncognito.com/oauth2/token",
+    default="EP_AUTH_URL",
```
NitishAroraKPMG, line 610: better to use the url
https://ep_auth.url

config/settings/base.py
ankitmahajan1: done

brahmaduttau (approved):

PR #1454, opened 2024-08-15
Hotfix/1.38.02 :: TEST deployment

pritish-ranjan01:


transit_odp/common/utils/s3_bucket_connection.py
```diff
@@ -101,11 +101,14 @@ def get_dqs_report_from_s3(report_filename):
...
     try:
+        bucket_name = getattr(
+            settings, "AWS_DATASET_MAINTENANCE_STORAGE_BUCKET_NAME", None
+        )
         storage = get_s3_bucket_storage()
 
-        if not storage.exists(csv_file_name):
-            logger.warning(f"{csv_file_name} does not exist in the S3 bucket.")
-            return [], None
+        storage.connection.meta.client.head_object(
```
PCridKPMG, line 109: Your head_object replacement doesn't fail gracefully if it fails to find the file - is this deliberate in that it'll Error if head_object fails?

pritish-ranjan01: Yeah, it will end up reporting 404 in the botocore.exceptions.ClientError exceptions

brahmaduttau (approved):

PR #1468, opened 2024-08-21
Story/bods 6387 fares publishing process - file level processing

surabhika:
1. Ability to handle ATCO codes when not an int value in the xml file 
2. Processing xml within a zip one at a time in order to reduce memory consumption
3. Skipping XML files that have failed XSD check from being processed and continue with processing remaining files
4. Generation of schema validation report for XML files that have failed schema validation
5. Fix number of trip and pass product count


ankitmahajan1 (approved):
approved

ndemenezes (approved):

PR #1476, opened 2024-08-23
fix avl tests

ljones92:


transit_odp/api/tests/test_avl_api.py
DeannaSharma, line 1: rather than passing through an empty object, I think we can remove the headers variable from `_get_consumer_api_response()` on line 121 in the `avl.py` file since its not being used now

DeannaSharma (approved):

PR #1487, opened 2024-08-27
Update agent_timetable_support.html

iMarkeh:
Removing my personal details from the BODS website. I have requested this twice previously but sadly it's still here.


ndemenezes:
@iMarkeh Closing this PR as it is being addressed in #1498 and will be incorporated in the next release to remove these details.

PR #1570, opened 2024-09-19
[BODS-6829] changes to update, insert, delete in chunks for naptan etl

ankitmahajan1:
This PR includes below changes 
1. updating the records in naptan tables (naptan_*) in chunks
2. instead of bulk update, update single record
3. update only specific records


amitkpmguk (approved):
looks good to me 👍

PR #1644, opened 2024-10-15
Fix/bods 7129

NitishAroraKPMG:


transit_odp/dqs/views/base.py
```diff
@@ -38,13 +38,16 @@ def is_dqs_new_report(self):
     def dispatch(self, request, *args, **kwargs):
         session_data = request.session
         self.show_suppressed = False
+        org_id = self.kwargs.get("pk1")
 
-        if session_data and session_data.get("_auth_user_id"):
+        if session_data and session_data.get("_auth_user_id") and org_id:
             auth_user_id = session_data.get("_auth_user_id")
             users = User.objects.filter(id=auth_user_id)
             if len(users) > 0:
```
pritish-ranjan01, line 46: Can just check if len(users):

transit_odp/common/templatetags/filters.py
```diff
@@ -26,6 +26,12 @@ def percentage(text, arg=-1):
     return floatformat(percentage_value, arg=arg) + "%"
 
 
+@register.filter()
+def subtract(value1, value2):
```
pritish-ranjan01, line 30: Do we need to handle negative numbers?

NitishAroraKPMG: we need the difference between two numbers

pritish-ranjan01 (dismissed):

pritish-ranjan01 (approved):

PR #1685, opened 2024-10-24
story/BODS-7071: Handle genuine clam AV rejections

ndemenezes:


transit_odp/validate/antivirus.py
```diff
@@ -96,23 +96,24 @@ def scan(self, file_: BinaryIO):
 
         if result.status == "ERROR":
             logger.info("Antivirus scan: FAILED")
-            raise AntiVirusError(file_.name)
+            raise AntiVirusError(file_.name, result.reason)
```
amitkpmguk, line 99: For AntiVirusError construct second parameter is line number, 
3rd parameter is message, 
Here we are passing message in line param

ndemenezes: Fixed this @amitkpmguk

amitkpmguk (approved):
Looks good to me 👍

PR #1701, opened 2024-10-28
BODS-7280 delete old observations for the same revision

surabhika:
Delete old schema and post schema violations even if the new file for the same revision does not contain any violations. This is to not show reports for the old files not present in the current version of the revision.


amitkpmguk (approved):
Looks good to me 👍

ndemenezes (approved):

PR #1708, opened 2024-10-29
BODS-7029 add new models to capture file level status

surabhika:
Screenshot 2024-10-29 at 12 17 34


transit_odp/pipelines/models.py
```diff
@@ -218,6 +219,104 @@ def handle_general_pipeline_exception(
...
+    error = models.CharField(max_length=255, choices=TASK_ERROR_CODE_CHOICES,unique=True)
+
+    class Meta:
+        db_table = "pipelines_pipelineerrorcode"
+    
+    def __str__(self):
+        return self.error
+
+
+class PipelineProcessingStep(models.Model):
```
surabhika, line 267: We might want to pre populate this table too?

transit_odp/pipelines/models.py
```diff
@@ -220,6 +221,113 @@ def update_progress(self, progress: int):
...
+        on_delete=models.CASCADE,
+    )
+    pipeline_error_code = models.ForeignKey(
+        PipelineErrorCode,
+        related_name="file_processing_errors",
+        on_delete=models.CASCADE,
+        null=True,  # Set to null to handle cases where no error occurs
+        blank=True,
+    )
+    created = models.DateTimeField(auto_now_add=True)  # Auto set the creation timestamp
```
andyecrow, line 312: I see in other places we're inheriting from `TimeStampedModel` to add `created` and `modified` fields - I guess we aren't doing that here because we don't care about the `modified` field?

andyecrow (approved):
Looks good to me!

transit_odp/pipelines/models.py
surabhika: Hi Andy, that made me think that we probably want to capture created and end date and time. So I've made those modifications

surabhika: I have finally used TaskResult as a base class for us to be able to have created, modified and completed times

ksrk (approved):

ksrk (approved):
Approved :)

surabhika:
Opening another PR

PR #1716, opened 2024-11-06
ignore unknown author branches

ankitmahajan1:


brahmaduttau (dismissed):

version.txt
robsteel-kpmg, line 1: This isn't required as it's not a change to the application itself, so no need to rebuild.

brahmaduttau (approved):

PR #1786, opened 2024-11-26
BODS-4476 remove unwanted annotations

surabhika:
The timetable list page used to have `BODS Compliance` label and a filter which were removed in the last release. The queryset hence pulled in data for `PTIObservations` and related tables and perform a few unwanted SQL joins. These are now removed along with queries to get admin areas too which has simplified the search query further. Here are the screenshots of timings before and after the change.
Before
Before1
before search

After
after
after search


brahmaduttau (dismissed):

brahmaduttau (approved):

ankitmahajan1 (approved):
approved

PR #1799, opened 2024-11-28
BODS-7555

SamerIssam:
Fix: Some journeys are not appearing for the correct line on the timetable visualiser


transit_odp/timetables/loaders.py
```diff
@@ -200,21 +200,24 @@ def load_services(self, revision):
...
                     service_patterns.reset_index()[
-                        ["service_pattern_id", "id", "file_id"]
+                        ["service_pattern_id", "id", "file_id", "line_name"]
                     ]
                     .drop_duplicates()
                     .rename(columns={"id": "id_service"})
                 )
+                # add line_name to the VJs
+                vehicle_journeys["line_name"] = (
+                    vehicle_journeys["line_ref"].str.split(":").str[-1]
```
alvak1, line 214: Please verify that line_ref will always have a value and of the expected format. str[-1] might throw an exception if semi colon is not found in the string

When possible it might be worth to merge lines to obtain line names

alvak1 (approved):
Approved

PR #1819, opened 2024-12-09
story/BODS-6189 + story/BODS-6187: Fix multiple requests to ABODS API and AVL requires attention column logic

ndemenezes:
Issues fixed:

- Logic behind the ‘AVL requires attention’ column is incorrect: Even if ‘AVL Published Status’ and ‘AVL to Timetable Match Status’ is both equal to Yes, ‘AVL requires attention’ still equals Yes instead of No.
- ABODS API number of request: For each line in the CSV that is being processed a request is sent to ABODS for a copy of the full dataset, which returns 10,644 lines. This is inefficient and causes delays (at times a gateway timeout error) when downloading the compliance report.


amitkpmguk (approved):
Looks good to me 👍

PR #1925, opened 2025-01-27
pin automake version to 1.16 to fix dependency issues

mTouhid:
`automake` has released version `1.17` today and `libgdal-dev` package has a dependency to version `1.16`.
Hence, pinning the `automake` version to `1.16`


mohit-kpmg (approved):
LGTM

PR #2009, opened 2025-02-20
Fix/bods 8118 Delete violations and txc attributes for revision update flow before triggering state machine

surabhika:


andyecrow (approved):
LGTM - I see the code to trigger the state machine (and delete the existing data) is duplicated between the views. We could extract it into a single "trigger state machine" function, but not a blocker 😄

PR #2026, opened 2025-02-27
Change logging level for missing tracks

SamerIssam:
A file does not have a tracks data will not be flagged as error.


pritish-ranjan01 (dismissed):

transit_odp/timetables/extract.py
```diff
@@ -357,7 +359,8 @@ def extract_journey_route_link(self):
         northing_easting_geometry = geo_type == "Easting/Northing"
         long_lat_geometry = geo_type == "Longitude/Latitude"
         if geo_type is None:
-            logger.error("Geometry type is not found")
+            logger.debug("Geometry type is not found")
```
NitishAroraKPMG, line 362: it will just suppress the alerting and should be added the correct logic

pritish-ranjan01 (approved):

PR #2042, opened 2025-03-04
fix/BODS-8401: Fix Overall Compliance Report - overall requires attention column logic

ndemenezes:
Fix for [BODS-8401](https://kpmgengineering.atlassian.net/browse/BODS-8401):

- When Scope status = “Out of Scope”, overall requires attention field is showing “Yes” instead of "No".
- Fares Timeliness Status is blank (instead of defaulting to "Not Stale") when Fares Published Status = "Unpublished".
- "NETEX:Last Modified Date" and "Date when fares data is over 1 year old" is showing values with timestamps instead of dates.


ankitmahajan1 (approved):
approved

PR #2062, opened 2025-03-10
fix/licence page: Fix Internal Server Error

ndemenezes:
Fix for internal server error showing up on DEV & TEST env when clicking on service number links.


amitkpmguk (approved):
Looks good to me 👍

PR #2114, opened 2025-03-31
fix/BODS-8138: Header Text Overflows Window Instead of Wrapping

ndemenezes:
https://kpmgengineering.atlassian.net/browse/BODS-8138


amitkpmguk (approved):
Looks good to me 👍

PR #2163, opened 2025-04-25
story/BODS-8568: Update PPC logic to handle where data for NOC and Line is for a different registration

ndemenezes:
Includes: https://kpmgengineering.atlassian.net/browse/BODS-8568


amitkpmguk (approved):
Looks good to me 👍

PR #2177, opened 2025-05-08
BODS-7438 decrease search term length to 2

surabhika:
This will affect operator search, LTA search, org search, and all other places where `AutoCompleteSearch` class is used.


amitkpmguk (approved):
Looks good to me 👍

PR #2217, opened 2025-06-05
BODS-9081 - Added a transmodel service pattern relation with tracks table

amitkpmguk:


andyecrow:
Something else we're going to need on this is the `sequence_number` - you can see how its specified on the existing `TracksVehicleJourney` model here: https://github.com/department-for-transport-BODS/bods/blob/dev/transit_odp/transmodel/models.py#L121-L129

andyecrow (commented):
LGTM

PR #2300, opened 2025-09-08
Fix queryset error in review view

surabhika:


amitkpmguk (approved):
Looks good to me :them

PR #2351, opened 2026-03-16
Removes conflicting dep source lists - previously unstable

HarrisonDuffield:
Resolves issue where there was conflicting dependencies.

https://github.com/department-for-transport-BODS/bods/actions/runs/23061505772/job/67199042035

See within above.

Steps Taken:
1) Remove debian unstable source list - unsure why this was here
2) Rebuild pipeline base image with no cache
3) Raise PR


prolude-dks (dismissed):
It appears that line adds the Debian unstable repository to APT sources during a Docker build, enabling installation of newer packages, but if added repeatedly it can produce duplicate repository warnings and dependency issues.

prolude-dks (approved):
Approved. Version changed.

PR #2358, opened 2026-04-20
DBODS-1337: optimise draft feed query and add concurrent dataset index

m-a-r-i-a-n:
What changed

- Removed the redundant organisation join from the draft feed raw SQL and filtered directly by organisation_id and dataset_type.
- Added a new non-atomic migration with a concurrent composite index on organisation_dataset (organisation_id, dataset_type).


a-joglekar (approved):
All changes look fine. Happy for you to merge

PR #2359, opened 2026-04-21
DBODS-1292: Add code to pull & store NaPTAN data in bucket

sophieoneill:
Pull Naptan data from API & upload to S3 bucket, Ticket: DBOS1292. 

- Updated the code so that instead of the Naptan data being pulled from the API and stored locally, the api is called and the resulting file is saved to S3. 
- It also archives the previous days file when a new file is added to the bucket.
- Data is in bucket bods-1297-data-landing-zone, in df-bods-nonprod


transit_odp/pipelines/pipelines/naptan_etl/extract.py
```diff
@@ -28,38 +33,109 @@
 logger = LoaderAdapter("NaPTANLoader", logger)
 
 
-def get_latest_naptan_xml():
-    naptan_url = settings.NAPTAN_IMPORT_URL
-    logger.info(f"Loading NaPTAN file from {naptan_url}.")
-    xml_file_path = None
+def get_naptan_s3_storage():
```
a-joglekar, line 48: Please remove all printf() statements, probably added to assist debugging

sophieoneill:
S3 bucket with file lablled "naptan_latest.xml"
image

Versioning in place
image

transit_odp/api/views/auth.py
```diff
@@ -0,0 +1,168 @@
...
+from rest_framework.response import Response
+from rest_framework.throttling import AnonRateThrottle
+from rest_framework.views import APIView
+
+
+class LoginRateThrottle(AnonRateThrottle):
+    """Throttle login attempts by IP to match allauth's rate limiting."""
+
+    def get_rate(self):
+        limit = getattr(settings, "ACCOUNT_LOGIN_ATTEMPTS_LIMIT", 5)
```
a-joglekar, line 29: Are these limit values for login reattempt and time out standard ? i.e. used same everywhere in system? Are they parameterised in other components, should there be requirement to change them? May be worth adding to parameters similar to S3 Bucket name, just in case we want to have ability to change them

transit_odp/api/views/auth.py
```diff
@@ -0,0 +1,168 @@
...
+class CurrentUserAPIView(APIView):
+    """Return the authenticated user from the session cookie."""
+
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        return Response(_serialize_user(request.user), status=status.HTTP_200_OK)
+
+
+class CurrentUserOrganisationsAPIView(APIView):
```
a-joglekar: Do we need this function? If code was reused and we don't need this method,  please remove

transit_odp/pipelines/pipelines/naptan_etl/extract.py
```diff
@@ -28,33 +33,153 @@
...
-    logger.info(f"Loading NaPTAN file from {naptan_url}.")
-    xml_file_path = None
+    verify_ssl = getattr(settings, "NAPTAN_SSL_VERIFY", True)
+    
+    logger.info(f"Loading NaPTAN file from {naptan_url} and saving to S3.")
+    #print(f"[naptan] Step 1: Connecting to DfT API at {naptan_url}...", flush=True)
 
     try:
-        response = requests.get(naptan_url)
+        response = requests.get(naptan_url, timeout=(30, 600), verify=verify_ssl, stream=True)
```
a-joglekar: Is this a standard value for timeout, as used in existing configuration? If possible, can we parameterise this as input to Lambda?

transit_odp/pipelines/pipelines/naptan_etl/lambda_handler.py
```diff
@@ -0,0 +1,61 @@
...
+            get_latest_naptan_to_s3,
+        )
+
+        naptan_key = get_latest_naptan_to_s3()
+        
+        return {
+            "statusCode": 200,
+            "body": json.dumps(
+                {
+                    "message": "NaPTAN archived successfully",
```
a-joglekar: Please change message to NaPTAN latest file downloaded successfully or something similar. We are doing away with explicit archival and using S3 versioning so please remove reference of archival related messages

transit_odp/api/views/auth.py
sophieoneill: Removed

a-joglekar (commented):
Changes look fine. Please merge the code

a-joglekar (approved):

HarrisonDuffield:
Unit tests ran on duplicate branch due to fork - https://github.com/department-for-transport-BODS/bods/actions/runs/25110034817/job/73581358502?pr=2373

PR #2365, opened 2026-04-22
reapplying DBODS-1337 changes

m-a-r-i-a-n:
*these were previously reverted for the release*

What changed
* Removed the redundant organisation join from the draft feed raw SQL and filtered directly by organisation_id and dataset_type.
* Added a new non-atomic migration with a concurrent composite index on organisation_dataset (organisation_id, dataset_type).


a-joglekar (commented):
Changes look fine. If you address Harrison's comment, this is good to be merged

PR #2371, opened 2026-04-27
DBODS-1357: precompute org AVL average for dataset export

m-a-r-i-a-n:
Moves org-level AVL average matching from per-row correlated SQL to a one-time precompute plus dictionary assignment in dataset export.


HarrisonDuffield:
Please up the version number :)

transit_odp/site_admin/exports.py
```diff
@@ -99,19 +110,44 @@ def get_ppc_weekly_per_feed_download_report(
...
+        .order_by("dataset_id", "-created")
+        .distinct("dataset_id")
+        .annotate(
+            percent_matching=Case(
+                When(
+                    vehicle_activities_analysed__gt=0,
+                    then=ExpressionWrapper(
+                        F("vehicle_activities_completely_matching")
+                        * 100.0
+                        / F("vehicle_activities_analysed"),
```
a-joglekar: Need to handle scenario where this variable is undefined or zero and you get divide by zero or undefined error. Applicable for both numerator and denominator variables of this calculation

m-a-r-i-a-n: AVL weekly average calculation is now explicitly handling null/zero PPC values (coalescing missing values and only calculating percentages when analysed count is greater than zero)

a-joglekar (approved):

PR #2372, opened 2026-04-27
DBODS:1293 Update code to read from the NaPTAN bucket

sophieoneill:
Updated the function get_latest_naptan_xml so that it pulls the Naptan data from the S3 bucket instead of from the API

Smoke test done locally 
```
from transit_odp.pipelines.pipelines.naptan_etl import extract as ex

key = ex.get_latest_naptan_to_s3()
print("uploaded key:", key)

xml_path = ex.get_latest_naptan_xml()
print("xml path:", xml_path)

df = ex.extract_stops(xml_path)
print("rows:", len(df))
print(df.head(3))
```
File is pulled as expected and stops are extracted as they should be
image


version.txt
```diff
@@ -1 +1 @@
-v1.54.23
+v1.54.23-2704
```
HarrisonDuffield: temporary version number used for deployment

.gitignore
```diff
@@ -143,6 +143,7 @@ $RECYCLE.BIN/
 .eggs/
 .env
 .env-prod
+.env.naptan-test
```
HarrisonDuffield, line 146: is this needeed

transit_odp/api/views/auth.py
```diff
@@ -0,0 +1,143 @@
+"""
+Session-based authentication API views for the Next.js frontend.
```
HarrisonDuffield, line 2: How did this get in here? ahah - quick revert or is this something that's needed here and this is a shortcut? 

If the latter can we take off the comment at the top here please :)

config/.env.local.template
```diff
@@ -116,6 +116,17 @@ ABODS_AVL_LINE_LEVEL_DETAILS_URL=
...
+# -------------------------------------------------------------------------------
+AWS_NAPTAN_RAW_STORAGE_BUCKET_NAME=
+NAPTAN_SSL_VERIFY=True
+
 # Support Contact Details
 SUPPORT_EMAIL=bodshelpdesk@kainos.com
 SUPPORT_PHONE=0800 000 000
+
+
+ACCOUNT_LOGIN_ATTEMPTS_LIMIT=5
```
HarrisonDuffield, line 129: This bit goes with the nextjs thing right?

HarrisonDuffield (changes requested):

transit_odp/pipelines/pipelines/naptan_etl/extract.py
```diff
@@ -28,38 +45,104 @@
...
+    s3_key = "raw/naptan/naptan_latest_xml.xml"
+    xml_file_path = None
+
+    try:
+        logger.info(f"Attempting to retrieve latest NaPTAN data from S3 at {s3_key}.")
+        if not storage.exists(s3_key):
+            logger.warning(f"No NaPTAN data found in S3 at {s3_key}")
+            return None
+
+        dir_path = Path(DISK_PATH_FOR_NAPTAN_FOLDER)
```
a-joglekar: There is no need to create a new copy of the file. Check if the directory and file exists. If so, return the existing file. Not sure what advantage we have of creating a new file.

We can reuse the following code:
for filename in os.listdir(DISK_PATH_FOR_NAPTAN_FOLDER):
                if filename.endswith(".xml"):
                    xml_file_path = os.path.join(DISK_PATH_FOR_NAPTAN_FOLDER, filename)

transit_odp/api/views/auth.py
sophieoneill: Oops, forgot to remove this one, gone now!

config/.env.local.template
sophieoneill: Yes, removed!

transit_odp/pipelines/tasks.py
```diff
@@ -22,9 +23,14 @@ def task_dqs_monitor():
     run_dqs_monitoring()
 
 
+@shared_task(ignore_result=True)
+def task_run_naptan_extract_etl():
+    naptan_extract_etl_main.run()
+
+
 @shared_task(ignore_result=True)
 def task_run_naptan_etl():
```
a-joglekar, line 34: Do we need to call this? I believe this should be removed completely as on extract ETL should be executed here

a-joglekar (changes requested):

transit_odp/pipelines/pipelines/naptan_etl/extract.py
```diff
@@ -13,53 +14,69 @@
...
-            with filepath.open("wb") as f:
-                for chunk in response.iter_content(CHUNK_SIZE):
-                    f.write(chunk)
+    try:
+        logger.info(f"Attempting to retrieve latest NaPTAN data from S3 at {s3_key}.")
+        if not storage.exists(s3_key):
+            logger.warning(f"No NaPTAN data found in S3 at {s3_key}")
+            return None
+
+        buffer = io.BytesIO()
```
a-joglekar: There is no need to create a new copy of the file. Check if the directory and file exists. If so, return the existing file. Not sure what advantage we have of creating a new file.

We can reuse the following code:
for filename in os.listdir(DISK_PATH_FOR_NAPTAN_FOLDER):
if filename.endswith(".xml"):
xml_file_path = os.path.join(DISK_PATH_FOR_NAPTAN_FOLDER, filename)

a-joglekar (changes requested):

sophieoneill:
The celery task for the naptan_extract_etl has not been set up yet as the schedule has yet to be defined. Task `task_run_naptan_etl()` was created for the celery run, to use when ready

transit_odp/pipelines/pipelines/naptan_etl/extract.py
```diff
@@ -13,53 +14,77 @@
...
+
+
 def get_latest_naptan_xml():
-    naptan_url = settings.NAPTAN_IMPORT_URL
-    logger.info(f"Loading NaPTAN file from {naptan_url}.")
-    xml_file_path = None
+    storage = get_naptan_s3_storage()
+    s3_key = "raw/naptan/naptan_latest_xml.xml"
 
+    if os.path.exists(DISK_PATH_FOR_NAPTAN_FOLDER):
```
a-joglekar: The code from line 55 to 82 needs to be removed and replaced with if the file (XML file exists at S3 bucket, construct the path to that file and return it. If no file exists, return none. We do not need to create a new copy of the file.

a-joglekar (approved):

HarrisonDuffield (approved):

HarrisonDuffield:
https://github.com/department-for-transport-BODS/bods/pull/2374 - actions ran on there

PR #2375, opened 2026-04-30
DBODS-1111: Update yarn.lock to resolve transitive axios to 1.15.2

neilobriaink:
Axios is pulled in via the existing swagger-ui dependency tree. The swagger-related subtree was refreshed in the lockfile to resolve axios to a compliant version. fast-xml-parser is not present in tracked dependency files.


HarrisonDuffield (changes requested):
can you bump the version file by 1 on the patch no please ?

(Just how this repo works - sorry should have mentioned)

HarrisonDuffield (approved):

PR #2378, opened 2026-05-11
DBODS-1389: add NPTG API ingestion to S3 landing bucket via extract pipeline

m-a-r-i-a-n:
Implements NPTG extract to fetch from the NPTG API and persist latest data to S3 at raw/nptg/nptg_latest.xml


a-joglekar (approved):
Changes look fine.

PR #2380, opened 2026-05-22
DBOS-1625 Add code to pull & store NOC data in bucket

CaelainnBrinkley:
Updating the code to all the NOC API to be called to download and store the dataset in and S3 bucket


transit_odp/pipelines/pipelines/naptan_extract_etl/extract.py
```diff
@@ -139,3 +139,52 @@ def get_latest_nptg_to_s3():
     except Exception as exc:
         logger.error("Exception while uploading NPTG data to S3.", exc_info=exc)
         raise
+
+
+def get_latest_noc_to_s3():
+    noc_url = settings.NOC_IMPORT_URL
+    latest_key = "raw/noc/noc_latest.xml"
+    verify_ssl = getattr(settings, "NAPTAN_SSL_VERIFY", True)
```
a-joglekar: Should reference specific variable for NOC. Please do not use NAPTAN_SSL_VERIFY env variable

transit_odp/pipelines/pipelines/naptan_extract_etl/extract.py
```diff
@@ -139,3 +139,52 @@ def get_latest_nptg_to_s3():
     except Exception as exc:
         logger.error("Exception while uploading NPTG data to S3.", exc_info=exc)
         raise
+
+
+def get_latest_noc_to_s3():
+    noc_url = settings.NOC_IMPORT_URL
+    latest_key = "raw/noc/noc_latest.xml"
```
a-joglekar: bods-transport-data-parser process needs NOC data in CSV format. You would need to get both XML and csv format data. The transport data parser only needs NOC Lines but other modules may need rest of tables.  Please refer to how it downloads CSV file in download_data function in here : bods-transport-data-parser/tree/main/src/transport_parser/external_data/noc_parser.py 

Disruptions need following tables (again in CSV format)
noclines, noc_table & public_name

Bods Intergated data 
Uses XML format data --> nocrecords.xml

ABODS ITM
Uses CSV Data --> NOC Code

Fares 
USES CSV Data --> NOCLines.csv, PublicName.csv, NOCLines.csv

transit_odp/pipelines/pipelines/naptan_extract_etl/extract.py
```diff
@@ -139,3 +139,52 @@ def get_latest_nptg_to_s3():
     except Exception as exc:
         logger.error("Exception while uploading NPTG data to S3.", exc_info=exc)
         raise
+
+
+def get_latest_noc_to_s3():
```
a-joglekar: The entire NOC code should be in its own directory parallel to NaPTAN Extract ETL folder. There is no dependency to merge this with Naptan extraction and should be a stand alone separate lambda

a-joglekar (changes requested):

config/settings/base.py
```diff
@@ -523,6 +524,36 @@
     "BANK_HOLIDAY_API_URL", default="https://www.gov.uk/bank-holidays.json"
 )
 
+# NOC import URL
+NOC_XML_IMPORT_URL = env(
+    "NOC_XML_IMPORT_URL",
+    default="https://www.travelinedata.org.uk/noc/api/1.0/nocrecords.xml",
+)
+
+domain = "https://www.travelinedata.org.uk"
```
a-joglekar: If this is specific to NOC data only, can we prefix with NOC_

config/settings/base.py
```diff
@@ -523,6 +524,36 @@
...
+endpoint = "/wp-content/themes/desktop/nocadvanced_download.php"
```
a-joglekar: If this is specific to NOC data only, can we prefix with NOC_

config/settings/base.py
```diff
@@ -523,6 +524,36 @@
...
+
+params = {
```
a-joglekar: If this is specific to NOC data only, can we prefix with NOC_

config/settings/base.py
```diff
@@ -523,6 +524,36 @@
...
+        "table_management_divisions",
+        "table_noclines",
+        "table_noc_table",
+        "table_operators",
+        "table_public_name",
+    ],
+    "submit": "Submit",
+}
+
+NOC_URL = f"{domain}{endpoint}?{urlencode(params, doseq=True)}"
```
a-joglekar: Please change the URL construction in line with change in parameter names (as per previous comments)

transit_odp/pipelines/pipelines/noc_extract_etl/extract.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if not csv_members:
+            raise ValueError("NOC ZIP archive did not contain any CSV files.")
+
+        for member in csv_members:
+            filename = os.path.basename(member)
+            if not filename:
+                continue
+
+            table_name = os.path.splitext(filename)[0]
```
a-joglekar: Before constructing destination path, please check if table_name is not null. It is possible that Traveline may add new tables to zip archive which we do not have mapping for. If such a situation occurs, log the name of the file which could not be mapped but iterate through rest of files.

a-joglekar (changes requested):

a-joglekar (approved):

PR #2402, opened 2026-06-17
Dbods 1638 - WECA Ingestion Pipeline

Aegman:
Adds data landing zone pipeline code for WECA data.

Pipeline will fetch both services and registrations datasets from WECA API (scheduled) and store in S3 bucket location.
Will create metadata JSON file, and will only write data to S3 if it has changed since the last check.

Should create the following files:
- weca_metadata_latest.json
- weca_registrations_latest.json
- weca_registrations_validatated_latest.json
- weca_services_latest.json
- weca_services_validated_latest.json


transit_odp/pipelines/pipelines/weca_extract_etl/client.py
```diff
@@ -0,0 +1,323 @@
...
+    start_point: str = Field(alias="startpoint_espt")
+    finish_point: str = Field(alias="endpoint_sp")
+    via: str = Field(alias="via_services_pt7atfu9e78z39yqc")
+    effective_date: date = Field(alias="proposedstartda_rviceslvicespt")
+    api_type: str = Field(default=API_TYPE_WECA)
+    atco_code: Optional[str] = Field(alias="fullserialnumbe_trationrations")
+
+    @field_validator("effective_date", mode="before")
+    def parse_effective_date(cls, value):
+        value = value.replace("Sept", "Sep")
```
a-joglekar, line 48: Do we need this? Looks like we have date format convertor and validator below? (Line # 180)

a-joglekar (changes requested):

transit_odp/pipelines/pipelines/weca_extract_etl/client.py
Aegman: Unfortunately yes, the date/time format only works with 3 character months. This was an issue raised previously: https://busopendataservice.atlassian.net/wiki/spaces/KBODS/pages/129040419/WECA+API+responses+sometimes+improperly+structured+causing+missing+or+incorrect+records+-+Fix+testing

a-joglekar (approved):

PR #2429, opened 2026-06-29
ci: DBODS-2045 update workflows relating to ECS to use AWS maintained actions

DeannaSharma:


.github/workflows/templates/update_task_def/action.yml
```diff
@@ -15,19 +16,22 @@ inputs:
...
+        container-name: ${{ inputs.container }}
+        image: ${{ inputs.image_id }}
+
+    - name: Deploy ECS task definition
+      uses: aws-actions/amazon-ecs-deploy-task-definition@a310a830f5c14e583e35d84e4e1ec7dd177c3c9c # 2.6.2
+      with:
+        task-definition: ${{ steps.render-task-def.outputs.task-definition }}
+        service: ${{ inputs.service }}
+        cluster: bodds-${{ inputs.environment }}
+        wait-for-service-stability: true
```
ljones92: I think by default this will use an exponential backoff so it could end up waiting a long time for a service that has reached stability. It might be worth explicitly setting `wait-max-delay-deconds` to 15 to get prompt feedback

ljones92: It's also worth noting that this change will make the deployment sequential, so it will update one service, wait for it to be stable, then update the next service and so on. Currently it updates all of the services and then ecs-wait will check them all at the same time. So this change will slow things down and could lead to a longer period of services running with different versions which could be problematic depending on the change. That will be happening now as well but at a smaller scale.

Some options are:

- Go with this approach and accept the increased skew time
- Set `wait-for-service-stability` to false and replicate ecs-wait with a simple bash script (this will mimic the existing flow so it's probably the safest approach)
- Use a matrix job and spin up a runner per service to do the upgrades in parallel
- Overhaul the deployment approach to use something like blue/green deployments but that's a much larger change

DeannaSharma: Thanks, have gone with the approach the bash script to replicate ecs-wait! Also updated to move these actions into a shared repo

.github/workflows/cavl_deployment.yml
```diff
@@ -105,53 +105,65 @@ jobs:
...
-        environment: ${{ env.ENVIRONMENT_NAME }}
+        task_definition_family: bodds_celeryflower_${{ env.ENVIRONMENT_NAME }}
         image_id: ${{ steps.check-image.outputs.image }}
+        cluster_name: bodds-${{ env.ENVIRONMENT_NAME }}
+        wait_for_service_stability: false
 
-    - name: Update celeryworker Task
-      uses: "./.github/workflows/templates/update_task_def"
+    - name: Update celeryworker task
+      uses: department-for-transport-BODS/.github/actions/update-ecs-task-definition@main
```
ljones92: It would be better to pin this to a commit SHA rather than use main. We should probably have a process to tag the shared actions especially if multiple repos are going to end up using the same ones.

ljones92 (approved):

PR #2435, opened 2026-07-06
fix: resolve swagger-ui react version to 18

ljones92:
- Resolve swagger-ui to use react and react-dom v18 to prevent version mismatch which is leading to the swagger UI not rendering
- Pin swagger-ui to v5.32.5 which is the last version to support node 20
- Add mise.toml for local tooling


HarrisonDuffield (changes requested):
version file but yes

HarrisonDuffield (approved):

PR #2443, opened 2026-07-13
DBODS-2143: Move to session auth jul

HarrisonDuffield:
- Moves publishing onto session-based auth w/ standardised pattern (CSRF + X-CSRFToken on mutating calls). Largely already in place
- Switches authenticated API fetching from `${config.djangoApiUrl}` prefixes to relative `/api/*` paths, forwarded to Django by the rewrite function in next.config.  Revisit at ALB/deployment time - is a better starting pattern to take it from there. 
- Removes now unnecessary endpoint-specific `route.ts` proxies that largely duplicated Django paths. AVL Archive path kept as was too bespoke - rest were very similar/boilerplate

Fares: 

See PR pointed at this one here <LINK>


bods-nextjs/app/publish/page.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    render(<PublishPage />);
+
+    await userEvent.click(screen.getByLabelText('Timetables'));
+    await userEvent.click(screen.getByRole('button', { name: 'Continue' }));
+
+    expect(getPaginated).not.toHaveBeenCalled();
+    expect(mockPush).toHaveBeenCalledWith('/publish/org/123/dataset/timetable');
+  });
+
+  it('routes agents with one organisation to organisation selection', async () => {
```
cpaul92, line 122: Should we add in a tests for users with multiple organisations? Essentially covering the default when it doesn't go in to either if

cpaul92: Also if we're adding tests can we add in checks for the expected url e.g. `/avl` vs `/timetables`. Same question for other test files

cpaul92 (commented):
lgtm, just a couple of additional test cases I think we can add

bods-nextjs/app/publish/page.test.tsx
HarrisonDuffield: Tests uploaded :)

cpaul92 (approved):

PR #2446, opened 2026-07-16
Dbods 831 - AVL Migration - Changes

neilobriaink:
Ticket: https://busopendataservice.atlassian.net/browse/DBODS-831

Follow up ticket to address issues found during demo as well as some other bits I found since.

Done:
More details page for service codes needing attention
Date/time format fix on AVL feeds list, feed page
Status info on avl feeds list
Default to asc on sort order avl feeds list
Implement Pagination correctly on avl feeds list/changelog
Add < back button to avl/123/delete
"Need help with operator data requirements?" under bar: single feed
Breadcrumbs trails and naming on various pages
Separate cancel page for update/new instead of modal
Draft feeds implementation 
Fix routing for cancel button when in draft status
Fix small ui differences
Fix autofill on update feed before published


bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/archive/page.tsx
```diff
@@ -15,6 +16,10 @@ function AvlArchivePageContent() {
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
 
+  useEffect(() => {
```
Comment (HarrisonDuffield) [inline bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/archive/page.tsx]: Is it possible to switch to this?
https://nextjs.org/docs/app/getting-started/metadata-and-og-images

bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/update/page.tsx
```diff
@@ -55,6 +56,8 @@ function AVLUpdatePageContent() {
   useEffect(() => {
     let cancelled = false;
 
+    document.title = 'Update data feed: upload';
```
Comment (HarrisonDuffield) [inline bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/update/page.tsx]: We should really be avoiding this I think. 

(some people do do it though i grant you but we can use the metadata)

Will also force a page refresh just to update the title I think

bods-nextjs/app/api/avl/consumer-feedback/[orgId]/route.ts
Comment (HarrisonDuffield) [inline bods-nextjs/app/api/avl/consumer-feedback/[orgId]/route.ts:1]: I removed alot of the route files in favour of a common proxy,  see if you still need them

HarrisonDuffield (changes requested):

bods-nextjs/app/api/avl/consumer-feedback/[orgId]/route.ts
Comment (neilobriaink) [reply bods-nextjs/app/api/avl/consumer-feedback/[orgId]/route.ts:1]: I've changed to use the common proxy where I could but kept ones needed for the report downloading

bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/archive/page.tsx
Comment (neilobriaink) [reply bods-nextjs/app/publish/org/[orgId]/dataset/avl/[datasetId]/archive/page.tsx]: Switched to this where possible for the static pages but have kept it in the case where there was an upload wizard and the title changes dynamically based on where it is in the process

bods-nextjs/app/govuk.scss
```diff
@@ -1 +1,7 @@
-@import "govuk-frontend/dist/govuk/index";
\ No newline at end of file
+@import "govuk-frontend/dist/govuk/index";
+
+@media (min-width: 40.0625em) {
```
HarrisonDuffield: where did this come from?

transit_odp/avl/views/api.py
```diff
@@ -229,6 +236,8 @@ def get_avl_review_status_api(request, pk1, pk):
             "shortDescription": revision.short_description,
             "comment": revision.comment,
             "urlLink": revision.url_link,
+            "username": revision.username,
+            "password": revision.password,
```
HarrisonDuffield: Not so sure we need to return the login details in the GET api?

HarrisonDuffield: as this is only being used for update then we just ask for the username/password at update time, even though we already have them server side

bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AvlAttentionContent.tsx
```diff
@@ -0,0 +1,243 @@
...
+                            />
+                          </g>
+                        </svg>
+                      </button>
+                    </div>
+                  </div>
+                </form>
+              </div>
+
+              {data?.noResults ? (
```
Comment (HarrisonDuffield) [inline bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AvlAttentionContent.tsx:169]: I think if an org doesn't need any attention it will also fall in here? 

(is this how it works on django - very possible)

___

If this is how the behaviour is on django then ignore

If not can we gate this to actually this so that this block only appears if a search content has been entered and then another for if nothing requires attention

Comment (HarrisonDuffield) [reply bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AvlAttentionContent.tsx:169]: transit_odp/avl/templates/avl/requires_attention.html - it actually might be pre-existing

bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+      router.back();
+      return;
+    }
+
+    globalThis.location.href = formUrl;
+  };
+
+  return (
+    <div className="govuk-width-container">
+      <Link type="button" className="govuk-back-link govuk-button-back-link" href={formUrl}>
```
Comment (HarrisonDuffield) [inline bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx:25]: Clicking back or cancel will work but I think it would take the user to a new state of page with nothing eneter?

I don't think anything is actually committed anywhere until the end so we'd need some way to keep state , either save it somewhere or pass it onto the cancel, or turn the cancel into a modal. 

Have a look how it was implemented in the django. If they were saving it when user enetered at every stage then we can't do that as we're API-ing to the end, but we could pass about into this component I would have thought - or build something on the side to assist with this

HarrisonDuffield (changes requested):
some more comments sorry

bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx
Comment (neilobriaink) [reply bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx:25]: On UAT, if you click cancel but then cancel that cancellation screen as it were, it brings you back to the upload wizard without any of the info the user had filled in before clicking cancel. I can implement keeping state here though if you think it's worth it - I feel like that should be the case and not make the user start again?

bods-nextjs/app/govuk.scss
neilobriaink: It was ported over from django while making the Pagination component but I'll move the class to globals.css and get rid of the media query as it doesn't seem to have an effect with other rules in place

bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AvlAttentionContent.tsx
Comment (neilobriaink) [reply bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AvlAttentionContent.tsx:169]: Yeah it's pre-existing I believe, it has the same message in both cases in django

bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx
Comment (HarrisonDuffield) [reply bods-nextjs/app/publish/org/[orgId]/dataset/avl/_components/AVLCreateCancelContent.tsx:25]: Yes that should be but it probably should be a bug ticket in that case and stop this scope creeping

HarrisonDuffield (approved):
Changes requested resolved and followedip

PR #2461, opened 2026-07-31
Dbods 1351 deploy bodsnextjs

HarrisonDuffield:
Changes Made:

- Dockerfile
- Github Actions
- CSRF related issues solved
- NextJS was using existing API for data retrieval in some places - which required auth it shouldn't have needed
- Merges Dev branch in


bods-nextjs/app/api/next/avl/archive/route.ts
```diff
@@ -1,43 +1,46 @@
 import { NextRequest, NextResponse } from 'next/server';
-import { config } from '@/config';
-import { getSessionHeaders, hasSessionCookie } from '../_utils/session-auth';
+import { getSessionHeaders, hasSessionCookie } from '@/lib/api-session';
+import { config } from '@/runtime-config';
 
 const DEACTIVATE_SUBMIT_BODY = 'submit=submit';
+const NUMERIC_ID = /^\d+$/;
+
+function isNumericId(value: string | null): value is string {
```
DeannaSharma: Might be worth moving this out to a utils folder as it looks like we do a similar check in the file above `bods-nextjs/app/api/avl/requires-attention/export/route.ts`

bods-nextjs/config/server.ts
```diff
@@ -0,0 +1,14 @@
+const djangoInternalOrigin = process.env.DJANGO_INTERNAL_ORIGIN || 'http://localhost:8000';
```
DeannaSharma, line 2:
```suggestion
import "server-only"

const djangoInternalOrigin = process.env.DJANGO_INTERNAL_ORIGIN || 'http://localhost:8000';
```

Worth adding this to make the build fail if any client component accidentally imports server-only config values

bods-nextjs/config/client.ts
DeannaSharma, line 1: Just a thought, config.ts vs runtime-config.ts file names could be a bit more descriptive to make it easy to understand the client vs server values.

We could always rename the files to make it explicit:
config.ts → client-config.ts (or config/client.ts)
runtime-config.ts → server-config.ts (or config/server.ts)

Might also be worth exporting the config objects as clientConfig and serverConfig so it's obvious which one you're using

.github/workflows/dev_deployment.yml
```diff
@@ -35,8 +43,19 @@ jobs:
...
+      with:
+        filters: |
+          django:
+           - version.txt
+          nextjs:
+           - bods-nextjs/package.json
+
     # Configure AWS credentials against the BODS Shared-Services account
     - name: Configure AWS credentials in Shared-Services
+      if: (github.ref == 'refs/heads/dev' && steps.deploy-changes.outputs.django == 'true') || (github.ref == 'refs/heads/bods-two' && steps.deploy-changes.outputs.nextjs == 'true') || (github.event_name == 'workflow_dispatch' && inputs.service == 'django') || (github.event_name == 'workflow_dispatch' && inputs.service == 'nextjs')
```
DeannaSharma:
```suggestion
```
I think we can remove this if condition because we always want it to run if we're doing a deployment? I'd also review some of the other actions as I think there's a few places this could be simplified :)

DeannaSharma: The conditionals in general make the workflow a little tricky to follow - something to consider is creating a new workflow files for just the nextjs deployment that runs on changes to the feature branch / when manually triggered? Then when the feature branch is merged we bring it into the normal deployment flows.

.github/workflows/dev_deployment.yml
```diff
@@ -35,8 +43,19 @@ jobs:
...
       uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # 6.0.3
 
+    - name: Detect service version changes
+      id: deploy-changes
+      uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # 4.0.1
+      with:
+        filters: |
+          django:
+           - version.txt
+          nextjs:
```
DeannaSharma: This might be a bit fragile as it won't capture all code changes.

If the intention is to update the version in the package.json to trigger a deployment then it might be worth adding a note to the README. Or we could always broaden the filter to `bods-nextjs/**` to capture any change to the new frontend code.

RhiannaReid (approved):

bods-nextjs/app/api/next/avl/archive/route.ts
HarrisonDuffield: Yeah done

bods-nextjs/config/client.ts
HarrisonDuffield: Yes good shout

bods-nextjs/config/server.ts
HarrisonDuffield: done

.github/workflows/dev_deployment.yml
HarrisonDuffield: I chucked the same check in everywhere stupidly yes. 

Simplified

(Would rather not complicate things further than we already have I think so will keep in the current scripts but simplified)

.github/workflows/dev_deployment.yml
HarrisonDuffield: Simplified above

DeannaSharma (approved):

PR #2476, opened 2026-08-13
Dbods 1531 again

HarrisonDuffield:
Works as a proxy rewrite to make things go to data. or publish dot rather than an approach of splitting into node zones.

Node Zone approach is probably overkill as we wouldn't benefit from the individual scaling it would provide. Unless we did envisage a future where we split the contianers into 3. (Also adds infra overhead to do this properly)

Majority of these changes are just file path changes type stuff


bods-nextjs/app/guidance/support/local-authorities/LocalAuthoritiesRequirementsPageContent.tsx
```diff
@@ -121,7 +122,7 @@ function LocalAuthoritiesContent() {
       <Breadcrumbs
         items={[
           { label: 'Bus Open Data Service', href: '/' },
-          { label: 'Publish Bus Open Data', href: '/publish' },
+          { label: 'Publish Bus Open Data', href: HOSTS.publish },
           { label: 'Guidance', href: '/guidance/support/bus-operators' },
```
DeannaSharma: Just wanting to check if this href is right? On prod going to `https://www.bus-data.dft.gov.uk/guidance/support/bus-operators` I get a 404 and other instances of this href use the wwwPath

HarrisonDuffield: It's not no - different issue but fixing before Phil finds that too 😆

bods-nextjs/proxy.ts
```diff
@@ -8,24 +15,125 @@ const subdomainRoutes: Record<string, string> = {
   admin: '/admin',
 };
 
+const wwwOnlyRoutePrefixes = [
+  '/accessibility',
+  '/changelog',
+  '/contact',
+  '/cookie',
+  '/guidance',
```
DeannaSharma: Do we need to add `/guidance/support/bus-operators` and `/guidance/support/developer` to this list also?

DeannaSharma:
Not sure if it's related to this code changes but noticed that the favicon isn't being applied on the main domain (`https://www.dev.bus-data.dft.gov.uk/`) but is applied on publish and data subdomains.

bods-nextjs/proxy.ts
HarrisonDuffield: THose are both incorrect similar to the above so will fix

DeannaSharma (approved):

PR #2487, opened 2026-08-27
Dbods 2771 fix routing

HarrisonDuffield:


bods-nextjs/app/publish/org/[orgId]/dataset/page.tsx
```diff
@@ -29,6 +29,8 @@ function SelectDatasetTypePageContent() {
       targetPath = `/publish/org/${orgId}/dataset/fares/create`;
     } else if (selectedType === 'avl') {
       targetPath = `/publish/org/${orgId}/dataset/avl/new`;
+    } else if (selectedType === 'timetable') {
+      targetPath = `/publish/org/${orgId}/dataset/timetable/create`;
```
Comment (neilobriaink) [inline bods-nextjs/app/publish/org/[orgId]/dataset/page.tsx]: If we only have 3 selectable types and 2 now land on /create, would it be better to check if it's avl, send it to /new if so, or else just default to /create?

```
if (selectedType === 'avl') {
      targetPath = `/publish/org/${orgId}/dataset/fares/new`;
} else {
  targetPath = `/publish/org/${orgId}/dataset/${selectedType}/create`;
}
```

Comment (HarrisonDuffield) [reply bods-nextjs/app/publish/org/[orgId]/dataset/page.tsx]: I am being stupid, none of this needs to be /create. That was an oddity of running real BODS and local BODS side by side

neilobriaink (approved):

PR #2488, opened 2026-08-27
Dbods 2642 account mgmt

HarrisonDuffield:
Changes made:

1) Account Management Pages
2) Account Handling reorganized into in effect own global service - Publish|Data|Account
3) Breadcrumb Changes
4) Password Reset Flow stuff


transit_odp/users/views/api.py
```diff
@@ -0,0 +1,896 @@
...
+    if dataset_type == DatasetType.FARES.value:
+        return "FARES"
+    return "TIMETABLE"
+
+
+def _subscription_status_display(
+    status: str | None, dataset_type: int
+) -> tuple[str, str]:
+    """Labels from organisation/snippets/status_indicator.html."""
+    if status == "live":
```
neilobriaink: Should these be an enum somewhere?

neilobriaink (approved):

transit_odp/users/views/api.py
HarrisonDuffield: Correct -applied

PR #2489, opened 2026-08-28
Widen inactive_service registraion_number 

tsoftly:
Increases registration number allowed size to account for new longer numbers


HarrisonDuffield (approved):
please update the version file to the next incremement

PR #2504, opened 2026-09-02
DBODS-2249

RhiannaReid:
added try/except around etree.parse in validate.py to fix the actual bug
added a helper to disable antivirus checks for testing (antivirus is automatically set to on unless value is changed from false to true in env)


config/settings/base.py
```diff
@@ -494,6 +494,10 @@
 # -----------------------------------------------------------------------------
 CLAMAV_HOST = env("CLAMAV_HOST", default="clamav")
 CLAMAV_PORT = env.int("CLAMAV_PORT", default=3310)
+# Local dev escape hatch: skip virus scan entirely (e.g. when ClamAV is rate-limited).
+DISABLE_ANTIVIRUS_CHECK = env.bool("DISABLE_ANTIVIRUS_CHECK", default=False)
```
HarrisonDuffield: Agree with this, can we put that as a separate PR from this change.

BODS doesn't run great locally without significant difficulty and we need to fix this; this can be the first step towards that easily enough but should go in separately to this

transit_odp/timetables/validate.py
```diff
@@ -74,23 +76,36 @@ def iter_get_files(self):
...
             error = XMLValidator(file_).dangerous_xml_check()
             if error:
                 violations.append(BaseSchemaViolation.from_error(error[0]))
                 continue
 
             file_.seek(0)
-            doc = etree.parse(file_)
+            try:
+                doc = etree.parse(file_)
+            except etree.XMLSyntaxError as exc:
```
HarrisonDuffield: We might aswell make this generic I think - most of the time yeah it'll be the XMLSyntaxError but any other error this lib throws will end up falling through the net in the same way it does now.

Unsure what they will be

transit_odp/timetables/validate.py
```diff
@@ -74,23 +76,36 @@ def iter_get_files(self):
...
+                violations.append(
+                    BaseSchemaViolation(
+                        filename=Path(name).name,
+                        line=exc.lineno or 0,
+                        details=exc.msg,
+                    )
+                )
+                self._failed_violation_filenames.append(name)
```
HarrisonDuffield: I've noticed this is called nowhere

transit_odp/timetables/validate.py
```diff
@@ -74,23 +76,36 @@ def iter_get_files(self):
...
-        for file_ in self.iter_get_files():
+        for name, file_ in self.iter_get_files():
             error = XMLValidator(file_).dangerous_xml_check()
             if error:
                 violations.append(BaseSchemaViolation.from_error(error[0]))
                 continue
 
             file_.seek(0)
-            doc = etree.parse(file_)
+            try:
```
HarrisonDuffield, line 87: would also put line 88 check inside this try just incase that first check fires a different exception - we can collect all in one place

HarrisonDuffield (changes requested):

HarrisonDuffield (approved):
Code looks good well done, probably remove those other files with just the empty line changes from it, and needs a reformat to get the actions to pass but other than that yes

Change version number too :)

Happy for self approve on fixing

PR #2517, opened 2026-09-11
Txc 2.4.1 change for hardcoding cleanup dropped branches

ekennedy-kainos:
This is to strip the hard-coded TXC value and introduce a variable to control the version

Jira: https://busopendataservice.atlassian.net/browse/DBODS-2931

this is for AC 1-3
The BODS frontend is update to state TransXChange 2.4.1 as pictured in the document below
The branches pulled out to aid the release are reapplied
Screens are as picutred in the Confluence doc below

AC 4 is on https://github.com/department-for-transport-BODS/bods/pull/2519


transit_odp/common/context_processors.py
```diff
@@ -53,9 +53,12 @@ def js_bundle_init(request):
 
 def global_settings(request):
     # return any necessary values
+    from transit_odp.timetables.constants import TXC_VERSION
```
HarrisonDuffield: Can this go to be a top level import in the file rather than local? 

Don't see anything circular to stop us

HarrisonDuffield (approved):

PR #2535, opened 2026-09-17
DBODS-2661: Timetable Upload Journey (FE2)

neilobriaink:
Ticket: https://busopendataservice.atlassian.net/browse/DBODS-2661

To match the timetable upload journey to the current frontend.

To note:
- I've removed a showShortDescriptionCounter prop that showed how many letters were remaining in the short description box. This doesn't happen in the current frontend. The prop also appeared in the AVL migration pages but was always set to false.
- Playwright tests added to 'e2e' folder for the timetable upload journey. This folder was originally made when doing the AVL migration so that we could redo the e2e tests in the same style as they appear in abods, instead of using the older 'e2e_tests' folder.


transit_odp/timetables/views/api.py
```diff
@@ -307,11 +440,73 @@ def get_timetables_review_status_api(request, pk1, pk):
...
+                "status": data_quality_status,
+                "score": data_quality_score,
+                "ragLevel": data_quality_rag.rag_level if data_quality_rag else None,
+                "criticalCount": critical_count,
+                "advisoryCount": advisory_count,
+                "reportUrl": data_quality_report_url,
+                "reportCsvUrl": data_quality_report_csv_url,
+                "showUpdate": show_update,
+            },
+            "ownerName": revision.dataset.organisation.name,
```
ekennedy-kainos, line 459: "ownerName": revision.dataset.organisation.name, already exists on line 437

transit_odp/timetables/views/api.py
```diff
@@ -256,6 +322,73 @@ def get_timetables_review_status_api(request, pk1, pk):
...
+            .filter(
+                status__in=[
+                    ReportStatus.REPORT_GENERATED.value,
+                    ReportStatus.REPORT_GENERATION_FAILED.value,
+                ]
+            )
+            .first()
+        )
+        if report:
+            data_quality_status = "SUCCESS"
```
ekennedy-kainos, line 361: just checking if this is expected behaviour
```
                status__in=[
                    ReportStatus.REPORT_GENERATED.value,
                    ReportStatus.REPORT_GENERATION_FAILED.value,
                ]
```

this means the data_quality_status will always be SUCCESS even with  ReportStatus.REPORT_GENERATION_FAILED.value right?

transit_odp/timetables/views/api.py
```diff
@@ -256,6 +322,73 @@ def get_timetables_review_status_api(request, pk1, pk):
     status = revision.status
     is_loading = _is_loading_status(status)
```
ekennedy-kainos, line 323: just a suggestion, but you got a is_loading variable which can be used to run queries once its done loading 
again not sure how much will be saved, feel free to auto resolve this 

something like: 
```
has_schema_violation = False
has_post_schema_violation = False
has_pti_observations = False
validation_report_url = f"/org/{pk1}/dataset/timetable/{pk}/review/pti-csv"

if not is_loading:
    has_schema_violation = SchemaViolation.objects.filter(
        revision=revision.id
    ).exists()
    has_post_schema_violation = PostSchemaViolation.objects.filter(
        revision=revision.id
    ).exists()
    has_pti_observations = PTIObservation.objects.filter(
        revision=revision.id
    ).exists()

```

loading at the rest of the function there's more that can be pulled into not is_loading

 txc_attributes 
metaData query
etc

transit_odp/timetables/views/api.py
neilobriaink: Yeah the current service has the same mapping, both report statuses are included in the query and any matching report sets it to success.

ekennedy-kainos (approved):

